PR #56, opened 2016-06-07
All requests to loolwsd are now prefixed with 'lool'

pranavk:
To be merged after https://gerrit.libreoffice.org/#/c/26029/


kendy:
Looking from my mobile, so I don't get why it should be necessary to add lool prefix in richdocuments? All should be transparent via hosting/discovery XML - right? Let's discuss on IRC before merging...

pranavk:
This is no longer needed now

PR #58, opened 2016-06-08
Don't load all documents when only one is clicked on from 'Files'

pranavk:


pranavk:
This fix doesn't work for shared documents. Better fix coming up ...

pranavk:
Updated the fix plus minor additions here and there.

the problem was, interestingly, sending fileId as string to getById for non-shared documents work, while it doesn't for shared documents, so i had to cast explicitly to int for all.

PR #64, opened 2016-06-14
version bump and same urlsrc for all docs

timar:


timar:
conflict, let's abandon this!

Issue #54, opened 2016-06-03
Please help me install richdocuments


BinhNguyen8x:
I've server Owncloud 9 on Ubutu14 , i don't install Richdocuments . Please all help me , please plase :((

timar:
richdocuments was developed for ownCloud 8.2. Reportedly it has problems on ownCloud 9. We will address this issue soon. You also need LibreOffice and LibreOffice Online (loolwsd). LibreOffice Online packages require systemd. Ubuntu 14 does not use systemd. So you need to write your System V init scripts (and ideally contribute them to the project).

BinhNguyen8x:
Thanks sir , sir can't guide me to install libreoffice on current owncloud

2016-06-03 14:30 GMT+07:00 Andras Timar <notifications@github.com>:

scolebrook:
@timar LibreOffice Online packages? From what I've seen it's just source code and you need to compile it yourself (including LibreOffice). Has this changed recently? If so, do you have any links to where packages can be obtained (even if still in alpha/beta/pre-release).

timar:
@scolebrook There are no LibreOffice Online binary packages, as far as I'm concerned. They can be built from source, or there is the updated [CODE VM from Collabora](https://www.collaboraoffice.com/code/) to play with, or there is the commercial offering from Collabora for hosters: [Collabora Online (COOL) v1.0 “Engine”](https://www.collaboraoffice.com/collabora-online-v1-engine/).

scolebrook:
@timar Thank's for the link to Collabora and their v1.0. I've been in contact with them since the start of the year but never saw anything regarding a 1.0. Great news.

Good software takes time but I still don't like the wait very much :smile:.

Issue #36, opened 2016-04-07
not previewURL ownclod9.0


shwarz13:
It does not show documents
create new documents , but does not show
in owncloud8.2.3 working

![- 07 04 2016 - 14 51 38](https://cloud.githubusercontent.com/assets/18259441/14350147/471f4950-fcd0-11e5-8f72-44eb4bdc76c2.png)
![- 07 04 2016 - 14 45 48](https://cloud.githubusercontent.com/assets/18259441/14350150/4ac70a5c-fcd0-11e5-8b3b-2d8960a6e083.png)

franklines:
Did you set up loolwsd and leaflet (server side of things)? RichDocuments is just the actual plugin for OwnCloud but you still have to setup the environment.

shwarz13:
owncloud.2.3, owncloud 9.0, loolwsd, leaflet installed on the same server.
loolwsd, leaflet used for both owncloud.
Client  (PC and browser) is the same.
owncloud8.2.3 it works.
owncloud9.0 error.
what I need to establish in the environment?

franklines:
@shwarz13 how did you compile loolwsd and leaflet? What directory did you use for the Libreoffice source tree?

shwarz13:
loolwsd

agree README
`    #libtoolize`
`    #aclocal`
`    #automake --add-missing`
`    #autoreconf`
`    #autoheader`

further I have collected a package for a Debian
`    #debuild --no-lintian -b`
install package

`    #loolwsd --version`
`    #1.6.0`

running
 `   #/usr/bin/loolwsd --systemplate=/opt/lool/systemplate --lotemplate=/opt/cp5.0 --childroot=/opt/lool/child-roots --numprespawns=1 `

loleaflet
I did
   ` #make`
and as in README

`    #jake build`
`    #cd plugins/draw-0.2.4`
`    #jake build  `  

LibreOffice/core
https://github.com/LibreOffice/core/tree/distro/collabora/cp-5.0

shwarz13:
thanks. everything is ok
OC 9.0.2

PR #66, opened 2016-06-15
Make me work with owncloud9

pranavk:
Due to API change (searchByMime), app was completely broken when
installed with owncloud9. This change should make it work with
owncloud9, while still keeping older owncloud version (upto 8.2)
in working state.

Lets use functions on object returned by searchByMime instead of
directly accessing its data members.


pranavk:
#54 #36

PR #67, opened 2016-06-15
Use new owncloud9 addMenuEntry plugin

pranavk:
... to add new odf files from main 'Files' app


timar:
What happens with ownCloud 8.2.x?

pranavk:
ownCloud 8.2.x doesn't support changing core from apps, so its not possible in versions older than ownCloud 9

PR #72, opened 2016-06-23
bccu#1882: Return to where user started by removing redundant call

pranavk:
Route /close/ doesn't mean anything, and we were making this
additional call always before closing the document. This had a
side-effect that sometimes /close/ call would be processed after
we assign returnToDir to window.location making browser always
stuck in richdocuments#index rather than reeturnToDir. This was
mainly the problem in firefox with owncloud9


timar:
The failed check was only a timeout of uploading code coverage, not a real problem.

PR #85, opened 2016-07-18
Name file info.xml directly

LukasReschke:
When developing apps for Nextcloud it is more usual that as a developer you can directly clone the app and enable it. The makefile is also very specific.

To improve the developer experience I have renamed that file to "info.xml" directly. At least I was kinda confused why I couldn't enable the app the first time I cloned it :-)

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -2,7 +2,7 @@
 <info>
 	<id>richdocuments</id>
 	<name>Collabora Online</name>
-	<description>Collabora Online allows you to to work with all kind office documents interactively direct in your browser.</description>
+	<description>Collabora Online allows you to to work with all kinds of office documents directly in your browser.</description>
 	<licence>AGPL</licence>
 	<version>1.1.0</version>
 	<author>Collabora Productivity based on work of Frank Karlitschek, Victor Dubiniuk</author>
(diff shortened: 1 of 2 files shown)


appinfo/info.xml
```diff
@@ -2,7 +2,7 @@
 <info>
 	<id>richdocuments</id>
 	<name>Collabora Online</name>
-	<description>An ownCloud app to work with office documents</description>
+	<description>Collabora Online allows you to to work with all kind office documents interactively direct in your browser.</description>
 	<licence>AGPL</licence>
 	<version>@PACKAGE_VERSION@</version>
```
DeepDiver1975: @PACKAGE_VERSION@ will be replaced during the build process as far as I can tell.
A simple rename will not work out

LukasReschke: Let's hardcode that then as well 😉

appinfo/info.xml
```diff
@@ -2,9 +2,9 @@
 <info>
 	<id>richdocuments</id>
 	<name>Collabora Online</name>
-	<description>An ownCloud app to work with office documents</description>
+	<description>Collabora Online allows you to to work with all kind office documents interactively direct in your browser.</description>
```
jancborchardt: Spelling corrected:
> Collabora Online allows you to to work with all kinds of office documents directly in your browser.

PR #86, opened 2016-07-18
Add support for opening and editing files from external storage

pranavk:


controller/documentcontroller.php
```diff
@@ -465,6 +471,9 @@ public function wopiGetFile($fileId){
 		//TODO: Support X-WOPIMaxExpectedSize header.
 		$res = $row->getPathForToken($fileId, $version, $token);
 
+		// Login the user to see his mount locations
+		$this->loginUser($res['owner']);
```
Ashod: No need to close the session like above?

lib/db/wopi.php
```diff
@@ -120,15 +120,6 @@ public function getPathForToken($fileId, $version, $token){
...
-		$view = new \OC\Files\View('/' . $owner . '/files');
-		$path = $row['path'];
-
-		if (!$view->is_file($path)) {
-			throw new \Exception('Invalid file path.');
-		}
-
-		$editor = $row['editor_uid'];
-		return array('owner' => $owner, 'editor' => $editor, 'path' => $path);
+		return array('owner' => $row['owner_uid'], 'editor' => $row['editor_uid'], 'path' => $row['path']);
```
Ashod: I guess we can skip validation here. Can't remember why I put it in the first place.

PR #88, opened 2016-07-18
Use event dispatcher

LukasReschke:
Registers the JS via event dispatchers. Otherwise the JS is not loaded when somebody has pretty URLs configured.


rullzer:
:+1:

PR #89, opened 2016-07-19
Use non strict comparison to validate token

icewind1991:
On my setup this is needed to load a document

Maybe related to using postgres


LukasReschke:
👍

PR #90, opened 2016-07-19
update name to fit in the app dropdown

jancborchardt:
The long title »Collabora Online« doesn’t fit in the app dropdown and is longer than all the other app titles. Also, »Online« is a bit obvious here. ;)
collabora title

Please review @timar @pranavk @karlitschek @LukasReschke – the main question is if it’s fine from a branding point of view.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <info>
 	<id>richdocuments</id>
-	<name>Collabora Online</name>
+	<name>Office</name>
 	<description>Collabora Online allows you to to work with all kinds of office documents directly in your browser.</description>
 	<licence>AGPL</licence>
 	<version>1.1.2</version>


LukasReschke:
cc @jospoortvliet @mmeeks Thoughts?

jospoortvliet:
Documents exist so Office is probably the best solution.

jancborchardt:
»Office« is good from my POV as well.

jancborchardt:
Yeah, »Office« fits best since it’s also a very descriptive name just like the other apps.

mmeeks:
Agreed; 'Office' is fine by me.
Thanks !

jancborchardt:
@mmeeks cool stuff, cheers!

karlitschek:
Nice

timar:
There is only one problem. This patch did not change the app name in the app dropdown. 

appinfo/app.php should have been patched instead

jancborchardt:
@timar ah right, missed to commit that. Thanks for fixing!

PR #173, opened 2017-05-11
Move fake login methods to Helper; use them before getting file content

pranavk:
In case server-side encryption is enabled, we need to make the user
login so as to load his/her private key before getting contents of the
file.


pranavk:
Working on a new solution.

PR #174, opened 2017-05-17
WOPI CheckFileInfo now returns a host-specific unique ID

Ashod:
This is used to uniquely identify a host even when
documents are accessed from alias hostnames.

### DIFF
--- a/controller/documentcontroller.php
+++ b/controller/documentcontroller.php
@@ -547,6 +547,9 @@ class DocumentController extends Controller {
 			return false;
 		}
 
+		$instanceId = \OCP\Config::getSystemValue('instanceid', '');
+		$instanceId = substr(md5($instanceId), -13); // Don't expose the secret ID.
+
 		$editorName = \OC::$server->getUserManager()->get($res['editor'])->getDisplayName();
 		return array(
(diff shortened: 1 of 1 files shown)


controller/documentcontroller.php
```diff
@@ -547,6 +547,9 @@ public function wopiCheckFileInfo($fileId){
 			return false;
 		}
 
+		$instanceId = \OCP\Config::getSystemValue('instanceid', '');
+		$instanceId = substr(md5($instanceId), -13); // Don't expose the secret ID.
```
pranavk: Secret is stored separately in the config file under 'secret' key. I don't understand the comment - "Dont expose the secret ID" here. We are just calculating a md5 hash of instanceId and calling the last 13 characters as secret ID ? Can't we just expose the secret ID directly here ? Please correct me if I am missing something.

Ashod: "secret" here refers to the instanceId. I assumed an admin wouldn't want to expose it, hence the md5.

pranavk: Okay, though I think there is no harm in exposing the InstanceId as it is, but no harm this way too.

Issue #175, opened 2017-05-17
Documents are not loading at all on 10.0.1RC1


tboerger:
I have tried to launch a demo, connected to a Collabora 2.0 instance. I have installed richdocuments through the marketplace (1.1.20). The example document is loading infinite. The only error I can see so far is a jQuery error within the browser console.

![bildschirmfoto 2017-05-17 um 12 01 28](https://cloud.githubusercontent.com/assets/156964/26148932/9e60e716-3af8-11e7-96b8-37fa97726f02.png)

pranavk:
Hmm, OC.Upload which was available to richdocuments pre-10.0.1 RC1 seems to be undefined now.

DeepDiver1975:
> Hmm, OC.Upload which was available to richdocuments pre-10.0.1 RC1 seems to be undefined now.

@PVince81 can you comment on this? THX

PVince81:
`OC.Upload` was renamed to `OC.Uploader`. In general it's not a public API.

There were some changes to that class which also required the gallery app to be changed.
In case it helps, see https://github.com/owncloud/gallery/pull/715/files#diff-57783a77dad9285b443294c9e7ed8d03R316

gklimm:
@tboerger You have to change apps/richdocuments/js/documents.js line 1065 OC.Upload._isReceivedSharedFile to OC.Uploader._isReceivedSharedFile

This work for me ...

pranavk:
Any thoughts on removing this app view from the app. One can always create new documents and upload from the files view anyway; this app view just adds to complexity without any advantage.

tboerger:
> @tboerger You have to change apps/richdocuments/js/documents.js line 1065 OC.Upload._isReceivedSharedFile to OC.Uploader._isReceivedSharedFile
> 
> This work for me ...

But to patch the files manually is not really a good solution :)

gklimm:
I'm agree with you but the only way to make it run for now.
I have the same opinion like pranavk. When the filelist was extended with the option to make new documents, the gallery view make no sense anymore. Gallery view for office documents doesn't make sense for me anytime :)

timar:
Fixed in 1.1.21 that I've just published on marketplace. We have not removed the app view yet, we just check for existence of OC.Upload and OC.Uploader.

timar:
Fixed in version 1.1.21.

PR #176, opened 2017-05-22
OC.Upload is now OC.Uploader in OC 10

pranavk:
We probably want to remove the app view completely in future. This is
just a quick hack.


pranavk:
#175

PR #182, opened 2017-06-06
Get file mtime before logging out

pranavk:
... otherwise, we return incorrect timestamp in PutFile response, which leads to some unexpected behavior in wsd.


pranavk:
@timar Important to get this in too. The external change patch pushed before collided regressed with encryption patch. This one should fix the regression.

PR #190, opened 2017-09-21
@NoCSRFRequired added

kapilnadiyapara:
@NoCSRFRequired check annotation added


pranavk:
Hi, as discussed, this doesn't sound like a good idea from security perspective. I have opened #191 to cater to your use-case.

timar:
Closing...

PR #196, opened 2017-11-08
Add support for PutRelativeFile for Save As.

kendy:
This needs Collabora Online 2.1.5 to show anything in the UI, but should be safe with 2.1.4 and older, and not affect those versions in any way.


pranavk:
Looks good to me. +1

PR #206, opened 2018-01-25
Propose filechange for app.php

T0mWz:
Adding an owncloud config option to hide the collabora menu icon from the OwnCloud App menu


pranavk:
Hi Tom,

This PR was good but we realized that using this option in config.php might not be a good idea. It is bit problematic in a clustered owncloud setup, I think, having to set the option in all the clusters manually. Instead, we moved the option to the UI which stores the value for this option in the database solving the multi-clustered problem as described above. 

#209 is the new PR for that. Sorry for not being able to merge this one. But really thanks for the idea. It's a good feature, no doubt.

PR #212, opened 2018-04-09
'save as' should save the file in editor's directory

pranavk:


pranavk:
The CI failures seem unrelated to me.

PR #224, opened 2018-10-21
Add note on Developer guidelines

mrow4a:
This PR adds note to the README on how to setup working development Collabora instance in few steps, without the need to have any advanced configurations/deployments relying on SSL configuration.

cc @DeepDiver1975 @VicDeo


timar:
Thanks, looks good, approved.

PR #236, opened 2018-11-30
Fix return to document from Collabora Editor

mrow4a:
Fixes https://github.com/owncloud/richdocuments/issues/238:
- return from normal and shared with user/group documents
- return from shared by link for both folder and file share

CC @pmaier1 @DeepDiver1975 @PVince81

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -616,6 +616,9 @@ var documentsMain = {
 		var dir = getURLParameter('dir');
 		if (dir != 'null')
 			documentsMain.returnToDir = dir;
+		var shareToken = getURLParameter('shareToken');
+		if (shareToken != 'null')
+			documentsMain.returnToShare = shareToken;
 
 		// this will launch the document with given fileId
 		documentsMain.show(documentsMain.fileId);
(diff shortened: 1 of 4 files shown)


mrow4a:
@DeepDiver1975 can we progress with this one?

js/documents.js
```diff
@@ -771,6 +774,9 @@ var documentsMain = {
 		if (documentsMain.returnToDir) {
 			documentsMain.overlay.documentOverlay('show');
 			window.location = OC.generateUrl('apps/files?dir={dir}', {dir: documentsMain.returnToDir}, {escape: false});
+		} else if (documentsMain.returnToShare) {
+			documentsMain.overlay.documentOverlay('show');
+			window.location = OC.generateUrl('s/{shareToken}', {shareToken: documentsMain.returnToShare}, {escape: false});
```
PVince81: in general, use OC.redirect() as it's mockable in JS tests.
I guess there are no JS tests here ?

js/viewer/viewer.js
```diff
@@ -188,7 +191,7 @@ $(document).ready(function() {
 		if (odfViewer.supportedMimes.indexOf(mimetype) !== -1 && $("#isPublic").val()){
 			// Single file public share, add button to allow view or edit
 			var button = document.createElement("a");
-			button.href = OC.generateUrl("apps/richdocuments/s/{token}?fileId={file_id}", { token: encodeURIComponent($("#sharingToken").val()), file_id: null });
+			button.href = OC.generateUrl("apps/richdocuments/public?fileId={file_id}&shareToken={shareToken}", { file_id: null, shareToken: encodeURIComponent($("#sharingToken").val()) });
```
PVince81: is this the kind of URL that people would be sending around ? what URL does the share dialog generate in this app ?

I'm only asking because changing the URL syntax could affect such users

mrow4a: This is internal API used by Collabora App JavaScript to redirect to Editor instead of downloading.

js/documents.js
mrow4a: There are no integration tests yet, yes.

PR #245, opened 2019-01-13
Add secure view using using IShare attributes

mrow4a:
- [x] Requires https://github.com/owncloud/core/pull/33994 to be finished in order to use extra permissions and "dav can-download" permission. 
- [x] Add admin setting to enable secure view in collabora
- [x] Make sure migration and compatibility with older versions is solved (old temporary token table is dropped)

secure-view-feature


js/viewer/extrapermissions.js
```diff
@@ -0,0 +1,29 @@
...
+		}
+
+		var mimetype = model.getFileInfo().getMimeType();
+		if (odfViewer.isSupportedMimeType(mimetype)) {
+			// With read-only permission set for the file, download will be disabled. Only viewing will be allowed
+			model.registerExtraSharePermission(
+				"dav",
+				"can-download",
+				true,
+				'allow download'
```
DeepDiver1975: translation?

js/viewer/extrapermissions.js
```diff
@@ -0,0 +1,29 @@
...
+			);
+
+			// With read-only permission set for the file, download will be disabled. Only viewing will be allowed
+			model.registerExtraSharePermission(
+				"richdocuments",
+				"can-print",
+				true,
+				'can print with watermark'
```
DeepDiver1975: translation?

lib/AppInfo/Application.php
```diff
@@ -119,6 +119,7 @@ public function registerScripts() {
 				function () {
 					\OCP\Util::addScript('richdocuments', 'viewer/viewer');
 					\OCP\Util::addStyle('richdocuments', 'viewer/odfviewer');
+					\OCP\Util::addScript('richdocuments', 'viewer/extrapermissions');
```
DeepDiver1975: only add if server supports it? add a check based on the server version maybe ....

mrow4a: @DeepDiver1975 do you have an example of app where such check is done? I cannot find any standard way of how to check server version

mrow4a: I also need to find a way of checking for version for "$share->getExtraSharePermissions".

js/viewer/attributes.js
```diff
@@ -0,0 +1,48 @@
...
+			throw 'missing OC.Share.ShareItemModel';
+		}
+
+		// Make can-download available permission as checkbox
+		var mimetype = model.getFileInfo().getMimeType();
+		var folderMimeType = 'httpd/unix-directory';
+		if (odfViewer.isSupportedMimeType(mimetype) || mimetype === folderMimeType) {
+			// With read-only permission set for the file, download will be disabled. Only viewing will be allowed
+			$.get(
+				OC.filePath('richdocuments', 'ajax', 'settings.php'),
```
DeepDiver1975: @mrow4a why is this call to the app needed? All share attributes should already be in the model - right?

PVince81: note: you can expose app settings using the config.php hook: https://github.com/owncloud/core/blob/stable10/core/js/config.php#L222 and an example here https://github.com/owncloud/core/blob/stable10/apps/files/appinfo/app.php#L65

js/admin.js
```diff
@@ -104,6 +104,37 @@ var documentsSettings = {
 		);
 	},
 
+	saveWatermarkText: function(value) {
+		$.post(
+			OC.filePath('richdocuments', 'ajax', 'admin.php'),
+			{ 'watermark_text': value }
+		);
+
+		OC.Notification.showTemporary(t('richdocuments', 'Saved watermark'), {timeout: 2});
```
DeepDiver1975: Watermark has been saved

lib/AppConfig.php
```diff
@@ -72,4 +74,18 @@ public function getUserValue($userId, $key) {
...
 		return $this->config->setAppValue($userId, $this->appName, $key, $value);
 	}
+
+	/**
+	 * Check if app can have enterprise features enabled
+	 *
+	 * @return bool
+	 */
+	public function enterpriseFeaturesEnabled() {
+		if ($this->config->getAppValue('enterprise_key', 'license-key', null) === null &&
```
PVince81: not enough, you need to also check whether the "enterprise_key" app is enabled which actually verifies the license key. if license key is empty or invalid, the app disables itself

js/viewer/attributes.js
```diff
@@ -0,0 +1,55 @@
...
+		if ((odfViewer.isSupportedMimeType(mimetype) || mimetype === folderMimeType) &&
+			OC.appConfig.richdocuments &&
+			OC.appConfig.richdocuments.secureViewAllowed &&
+			OC.appConfig.richdocuments.defaultShareAttributes) {
+			/** @type OC.Share.Types.RegisteredShareAttribute **/
+			var secureViewCanDownload = {
+				"scope": "permissions",
+				"key": "download",
+				"default": JSON.parse(OC.appConfig.richdocuments.defaultShareAttributes.secureViewCanDownload),
+				"label": t('richdocuments', 'can view without watermark'),
```
PVince81: please change to "can view/download (without watermarks)"

js/viewer/attributes.js
```diff
@@ -0,0 +1,55 @@
...
+				"incompatibleAttributes": []
+			};
+			model.registerShareAttribute(secureViewCanDownload);
+
+			/** @type OC.Share.Types.RegisteredShareAttribute **/
+			var secureViewCanPrint = {
+				"scope": "richdocuments",
+				"key": "print",
+				"default": JSON.parse(OC.appConfig.richdocuments.defaultShareAttributes.secureViewCanPrint),
+				"label": t('richdocuments', 'can print/export (will include watermark)'),
```
PVince81: please change to "can print/export (with watermarks)"

js/viewer/attributes.js
```diff
@@ -0,0 +1,55 @@
...
+			OC.appConfig.richdocuments &&
+			OC.appConfig.richdocuments.secureViewAllowed &&
+			OC.appConfig.richdocuments.defaultShareAttributes) {
+			/** @type OC.Share.Types.RegisteredShareAttribute **/
+			var secureViewCanDownload = {
+				"scope": "permissions",
+				"key": "download",
+				"default": JSON.parse(OC.appConfig.richdocuments.defaultShareAttributes.secureViewCanDownload),
+				"label": t('richdocuments', 'can view without watermark'),
+				"description": t('richdocuments', 'With this permission being' +
```
PVince81: from what I see the "description" field is not used in the core PR, so please remove it.

js/documents.js
```diff
@@ -287,12 +293,12 @@ var documentsMain = {
...
 			var fileName = documentsMain.fileName.substring(0, documentsMain.fileName.indexOf('.'));
 			var downloadUrl, restoreUrl;
-			// Tweak the fileId format as {fileId_instanceId_version}. this is what WOPI backend expects the fileId to be in
-			// Ofc, if no version, then just {fileId_instanceId}
+
 			if (version === 0) {
 				formattedTimestamp = t('richdocuments', 'Latest revision');
 				downloadUrl = OC.generateUrl('apps/files/download'+ documentPath);
 			} else {
+				// FIXME: this is no longer supported in OC10
```
PVince81: fix this ?

js/documents.js
```diff
@@ -311,12 +317,14 @@ var documentsMain = {
...
 			html = $(html).attr('data-fileid', fileId)
-				          .attr('data-title', fileName + ' - ' + formattedTimestamp);
+				.attr('data-version', version)
+				.attr('data-title', fileName + ' - ' + formattedTimestamp);
 			$('#revisionsContainer ul').append(html);
 		},
 
 		fetchAndFillRevisions: function(documentPath) {
 			// fill #rev-history with file versions
+			// FIXME: this is no longer supported in OC10
```
PVince81: fix this ?

lib/Controller/DocumentController.php
```diff
@@ -763,19 +834,50 @@ public function wopiCheckFileInfo($fileId) {
...
+			$canDownload = $res['attributes'] & WOPI::ATTR_CAN_DOWNLOAD;
+			if (!$canDownload) {
+				$watermark = \str_replace(
+					'{viewer-email}',
+					$editor->getEMailAddress() === null ? $editor->getDisplayName() : $editor->getEMailAddress(),
+					\OC::$server->getConfig()->getAppValue('richdocuments', 'watermark_text', '')
+				);
+				$result = \array_merge($result, [
+					'WatermarkText' => $watermark,
+					'DisableExport' => true, // FIXME: new permission?
```
PVince81: no need for "FIXME". maybe raise a ticket for future ?

templates/admin.php
```diff
@@ -65,4 +65,31 @@
...
+} else {
+	p(' />');
+}?>
+	<label for="enable_secure_view_option_cb-richdocuments"><?php p($l->t('Enable Secure View for read-only shares (requires enterprise version)')) ?></label>
+	<div id="enable-share-attributes-defaults" style="padding-left: 28px;" class="indent <?php if ($_['secure_view_option'] !== 'true') {
+	p('hidden');
+} ?>" >
+		<p style="max-width: 50em;"><em><?php p($l->t('Set default share permissions for all users')) ?></em></p>
+		<input type="checkbox" id="secure_view_can_download_default_option_cb-richdocuments" <?php p($_['secure_view_can_download_default'] === 'true' ? 'checked' : '') ?> />
+		<label for="secure_view_can_download_default_option_cb-richdocuments"><?php p($l->t('can view with watermark (disabling allows only viewing with watermark)')) ?></label>
```
PVince81: please use the same text as for the checkboxes in the share dialog

cc @pmaier1

appinfo/Migrations/Version20190310162809.php
```diff
@@ -0,0 +1,101 @@
...
+		]);
+
+		// Expiration time of the token
+		$table->addColumn('expiry', 'integer', [
+			'length' => 4,
+			'unsigned' => true,
+			'notnull' => true,
+		]);
+
+		$table->addUniqueIndex(['token'], 'richdocuments_wopi_token_idx');
```
PVince81: @mrow4a did you diff with the old DB schema at least with MariaDB ? from my experience it's easy for discrepancies to sneak in

appinfo/Migrations/Version20190310162809.php
```diff
@@ -0,0 +1,101 @@
...
+/**
+ * Auto-generated migration step: Please modify to your needs!
+ */
+class Version20190310162809 implements ISchemaMigration {
+	public function changeSchema(Schema $schema, array $options) {
+		$prefix = $options['tablePrefix'];
+
+		// clean old legacy table if exists (only temporary tokens)
+		if ($schema->hasTable("{$prefix}richdocuments_wopi")) {
+			$schema->dropTable("{$prefix}richdocuments_wopi");
```
mrow4a: @PVince81 I drop the table here. So indexes will also go away I think.

appinfo/Migrations/Version20190310162809.php
mrow4a: I drop table, and add only entries which are required

js/documents.js
mrow4a: Not scope of this PR

js/documents.js
mrow4a: Not a scope of this PR

lib/AppConfig.php
mrow4a: I think this is only about soft limit - with invalid key supplied, proper notification will be shown

PVince81: for all other apps the apps will auto-disable if the key is invalid, so here I think the feature should be disabled as well.

@pmaier1 or was this on purpose ?

mrow4a: @PVince81 the feature will be disabled without "any" key - valid or invalid. If someone puts invalid, ownCloud will scream that invalid key is supplied

PVince81: an invalid key will auto-disable the enterprise_key and also auto-disable any enterprise apps, no screaming.

mrow4a: @PVince81 I can do whatever action is needed, can you point me? I personally would prefer just a soft limit.

mrow4a:
@PVince81 can you recheck? shall we have some QA by product team?

js/admin.js
```diff
@@ -104,6 +104,37 @@ var documentsSettings = {
...
+	},
+
+	saveSecureViewOption: function(value) {
+		// FIXME: if version too low, throw error that this permission cannot be set for version <10.2 ?
```
PVince81: another alternative is setting min-version to 10.2 for the app, but 10.2 isn't out yet. we could merge this later

PVince81: or is there proper error handling on the UI in case this call failed ?

you can also add a version check, see in the `OC` namespace you might be able to find the current OC version...

lib/AppConfig.php
PVince81: I can't find anything in our discussions about the desired approach.

I'd prefer to do the check like we do it everywhere else, check that the app is enabled.

js/admin.js
mrow4a: @PVince81 hmm - true will check compatibility with earlier version. yes, lets merge it after 10.2 is out.

lib/AppConfig.php
pmaier1: Agree with @PVince81. The feature (in settings as well as in the sharing dialog and backend) should only be available when `enterprise_key` is enabled. Using the same logic as other EE apps is absolutely fine. We accept that it's easy to circumvent the restriction if you know what you're doing.

js/admin.js
mrow4a: it is already done (min 10.2)

lib/AppConfig.php
PVince81: solved now

mrow4a:
@PVince81 @micbar Tested on oc10.2.0RC2:
- upgrade from Colabora 2.1.2 to SecView (2.2.0) and all works smooth and temp table gets recreated with new schema
- All required featureset is working

mrow4a:
@PVince81 @pmaier1 can we consider this PR ready to merge?

PVince81:
I consider this ready to merge.

Not sure if we should wait for approval from @timar

pmaier1:
Please merge this. Then we can do translations and release officially. THX!

mrow4a:
@pmaier1 please consult @timar and @PVince81 for merging

PVince81:
- [ ] BUG: clicking on "Enable secure view" text in the settings doesn't tick checkbox. need to use the "label" tag properly. it works for other options. (can be fixed in a separate PR)

Issue #233, opened 2018-11-22
Open documents in a new tab


Rudloff:
Hello,

We are using richdocuments 2.0.9 with ownCloud 9.1.8

Some of our users are complaining that documents are always opened in the current tab, which is not very good for multitasking (if you want to open a couple of documents before starting to work, for example).
It is not even possible to do this manually, because *right click > open in a new tab* downloads the document instead.

Would it be possible to add an option to open documents in a new tab?

PR #248, opened 2019-01-25
Added Open documents in a new tab

merttumer:
https://github.com/owncloud/richdocuments/issues/233


mrow4a:
@merttumer can you repush, for some reason checks got stuck

PR #252, opened 2019-02-13
Fix CI Makefile issues

phil-davis:
1) The ``owncloud-coding-standard`` pipeline step in ``.drone.yml`` was duplicated - fix that
2) Add vendor and vendor-bin to clean-deps in Makefile so that everything can be cleaned
3) Define the ``all_src`` and ``dist_dir`` vars which are used later in ``make dist``
4) Add the "standard" "signing" block of code to ``Makefile``
5) Include ``admin.php`` and ``settings.php`` in the dist tarball


Makefile
```diff
@@ -14,7 +14,11 @@ NODE_PREFIX=$(shell pwd)
 
 app_name=richdocuments
 project_dir=$(CURDIR)/../$(app_name)
-build_dir=$(CURDIR)/build/artifacts
+doc_files=README.md
```
phil-davis: In the root folder there is also ``admin.php`` and ``settings.php``
What are those there for?
Should they be listed somewhere here so that they end up in the tarball?

phil-davis:
Also, why does ``Makefile`` have the ``appstore:`` target?
That looks like it does similar to ``make dist`` but in a different way.

Makefile
PVince81: These are the old ways how settings panels used to be registered.

I checked info.xml and it seems it wasn't ported to the new way, so these files are required I'd say... seems this wasn't discovered yet

PVince81:
@phil-davis I've seen many other Makefiles with a similar pattern.

phil-davis:
@PVince81 this is ready to review again.
When we decide a "new standard" ``Makefile`` format/infrastructure then we can cleanup whatever else is "old rubbish".

PVince81:
@timar FYI, not sure if you used the Makefile for past releases ?

PVince81:
anyway, moving forward for now, it was broken anyway...

PR #253, opened 2019-03-04
[poc] Use common makefiles

phil-davis:


micbar:
@phil-davis Did we drop this PR? OR is the refactoring ongoing?

phil-davis:
We decided not to go this way, but to put common makefile code blocks somewhere else (in another repo like done with owncloud-codestyle?).

I had left these PRs open so that I remember them. But I will close them now - the branches will still be there when I need to look at examples of what I was doing.

PR #259, opened 2019-06-15
Dont enable incognito mode if no encryption enabled

mrow4a:
Currently with encryption enabled, app requires incognito mode to be enabled. Without the encryption, the incognito causes that activity for this app is not auditable.

Future work: make collabora work with audit for master key encryption (for user key we cannot do anything unfortunetely)


mrow4a:
Obsoleted by https://github.com/owncloud/richdocuments/pull/260

PR #264, opened 2019-07-16
Impr sec view

mrow4a:
Logic for custom share attributes in javascript has been moved to an app. 

- now apps use only "core api" to set attributes, but the checkboxes logic stays in the app
- admin panel does not include "default share attributes", as secure-view checkbox is always visible (it is also technicaly difficult to set default attribute enabled during share creation)


micbar:
Version constraint has blocked CI.
Stable10 is now 10.3, please try again.

js/viewer/shareoptions.js
```diff
@@ -0,0 +1,488 @@
...
+			// determine if checkbox is enabled
+			// looping over checkbox defined and currently set attributes
+			var enabled = null;
+			for(var chIndex = 0; chIndex < checkboxAttributes.length; chIndex++) {
+				var checkboxAttribute = checkboxAttributes[chIndex];
+
+				for(var attrIndex = 0; attrIndex < attributes.length; attrIndex++) {
+					var currentAttribute = attributes[attrIndex];
+
+					if (currentAttribute.scope === checkboxAttribute.scope
```
PVince81: do we really need the "scope" property still or could a lot of these loops be implemented in a simpler way, for example by having the click handlers take care of exclusions in a hard-coded way ? (hard-coded would be fine as it's only 3 properties, if it helps with having less code and more readability)

mrow4a: @PVince81 yes, scope is needed as this allows to resolve conflicts between apps on attributes. With app/core scope it is highly unlikely that someone writes 2 checkboxes for 2 different apps using the same 1 attribute. 

Regarding readability, this is true that it could just be hardcoded. Should I take care?

PVince81: I'm mostly considering that other apps might want to copy the code for this approach, and this is a lot of code. If you think that hard-coding more would significantly reduce the amount of code to copy, then go for it (separate PR?).

Otherwise let's leave this.

js/viewer/shareoptions.js
```diff
@@ -0,0 +1,416 @@
...
+					});
+
+					// print can only be set when secure-view is enabled
+					if (download.enabled === false && watermark.enabled === true && print !== null) {
+						shareOptionsData.push({
+							cid: view.cid,
+							shareId: share.id,
+							shareWith: share.share_with,
+							name: "secure-view-print",
+							label: "can print / export",
```
PVince81: translations for labels ?

mrow4a:
@micbar @PVince81 acceptance tests adjusted, ready to rereview

individual-it:
tests look ok

PR #269, opened 2019-07-30
Remove stable10 test jobs

phil-davis:
Issue https://github.com/owncloud/core/issues/35777

The old core `stable10` is now `master`

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -237,53 +237,14 @@ matrix:
       NEED_INSTALL_APP: true
 
     - PHP_VERSION: 7.0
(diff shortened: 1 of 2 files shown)


phil-davis:
ToDo: look into why an acceptance test step definition has gone missing:
```
--- FeatureContext has missing steps. Define them with these snippets:

    /**
     * @When the user sets the sharing permissions of :arg1 for :arg2 using the webUI to
     */
    public function theUserSetsTheSharingPermissionsOfForUsingTheWebuiTo($arg1, $arg2, TableNode $table)
    {
        throw new PendingException();
    }
```

PR #270, opened 2019-07-31
distinguish between user and group while setting sharing permissions

HariBhandari07:
PR https://github.com/owncloud/core/pull/35944 changed the step:
```And the user sets the sharing permissions of "User One" for "simple-folder" using the webUI to```
to
```And the user sets the sharing permissions of user "User One" for "simple-folder" using the webUI to```
to distinguish between the user and group share permissions.
This pr updates ```richdocuments``` to the changes made in core.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -237,53 +237,14 @@ matrix:
       NEED_INSTALL_APP: true
 
     - PHP_VERSION: 7.0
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: this ended up happening in PR #269 because both this change and the core stable10->master change needed to happen together to get CI to pass.

PR #281, opened 2019-09-30
Update info.xml

timar:
Note: it is possible to use Collabora Online’s integration with re-compiled and/or re-branded backends.


timar:
No idea how to merge this. Why the codecov/patch and codecov/project were not running?

phil-davis:
codecov is a PITA - quite often the result does not get processed, or the status does not get back to GitHub. I used the GitHub API to post "success results" for codecov here - this PR does not effect code anyway.

Issue #289, opened 2019-10-02
Typo in Encryption warning message


davitol:
<img width="787" alt="Screen Shot 2019-10-02 at 10 11 21" src="https://user-images.githubusercontent.com/10596560/66028398-4e1ea200-e4fd-11e9-8b7e-ddfa893c40bf.png">

davitol:
Not only the typo in 'priviledged' but also why the message refers to 'user key is enabled' when the Default encryption module i have used is Master key. 


<img width="679" alt="Screen Shot 2019-10-02 at 10 15 39" src="https://user-images.githubusercontent.com/10596560/66028708-f2a0e400-e4fd-11e9-8ee8-db8e0e773091.png">

(@mrow4a jFYI)

mrow4a:
@davitol thanks, I will have a look - the message indeed should not be displayed and this was overlooked.

mrow4a:
@pmaier1 @micbar @davitol fix in https://github.com/owncloud/richdocuments/pull/292

PR #260, opened 2019-06-23
Make master encryption compatible with audit/activity

mrow4a:
There were few problem with previous implementation:
- with encryption or without, all activity was not auditable (incognito mode by default)
- public link was set for editor as owner of the file
- non-uniform method for getting priviledged access to file from collabora
- it was not made clear to the admin that this app is "partially" compatible with user key encryption as this app requires priviledged access to the files (and password of user with which files are encrypted cannot retrieved)

This PR fixes it:

coll2
coll1

Tested:
- [x] [without encryption] edit as owner, edit as share receiver, edit as public link, move files around as share receiver, ensure activity is registered, ensure versions are correctly generated
- [x] [with master key encryption] edit as owner, edit as share receiver, edit as public link, move files around as share receiver, ensure activity is registered, ensure versions are correctly generated
- [x] [with user key encryption] settings page warning displayed, coediting already opened file, edit file from root folder (no moving around), public links work, ensure versions are correctly generated

PR #292, opened 2019-10-08
[289] Fix app settings encryption message

mrow4a:
Fixes https://github.com/owncloud/richdocuments/issues/289


jvillafanez:
I don't like the dependency with the encryption app module... any other alternative?

In addition, I think you're assuming that the app module will be enabled, what if there a different module enabled?

mrow4a:
@jvillafanez any other ideas on how to achieve it? I could not find any other (actually this is copied from other place)

jvillafanez:
> @jvillafanez any other ideas on how to achieve it? I could not find any other (actually this is copied from other place)

I've asked first! :sweat_smile: 

I can overlook this since it's copied, but I have to complain this isn't a good solution.

Why do we need to check for encryption in the first place? Why should the richdocument app care about encryption being enabled or not?

mrow4a:
@jvillafanez this is related to the fact that Collabora/OnlyOffice/etc use "privileged" access to the files as they act on behalf of the user throught API - it means you need to login as editor (in case of normal and masterkey) or user incognito access (user key) - https://github.com/owncloud/richdocuments/blob/master/lib/Controller/DocumentController.php#L1084-L1119

some more read - https://github.com/owncloud/richdocuments/pull/260

jvillafanez:
Ok, but this is a similar case that happens with the sharing, isn't it?
If userA shares a file with userB so he can edit the file, and the file is encrypted. I think userB can edit the file regardless of how it's encrypted (I might be wrong)

mrow4a:
@jvillafanez it is about activity app in this issue https://github.com/owncloud/enterprise/issues/2619 . In user encryption I think sharing is even more complicated than that even with incognito mode.

mrow4a:
@PVince81 is there a prettier way of checking for master/user encryption?

PVince81:
ask @sharidas ^

sharidas:
Check for user keys encryption or masterkey encryption finally reaches down to the appconfig.
So looking into the pr, the check for encryption enable looks ok to me, also the check for masterkey encryption looks ok to me. Regarding app dependency, I don't have insight over this app.

jvillafanez:
If we're under time constraints I guess we'll have to use this fix. However, I think we should find a better solution.

I think core should somehow provide an API to transparently encrypt and decrypt the file regardless of what encryption app / mode / algorithm is used. This app shouldn't even care that the file will be encrypted. The FS should hide this complexity.

> @jvillafanez this is related to the fact that Collabora/OnlyOffice/etc use "privileged" access to the files as they act on behalf of the user throught API 

I still don't understand why. If I login a userA and edit a file, assuming that it will be collabora the one saving the file, it should send that userA is the one editing the file, and ownCloud should login with userA. For there, you just save the file normally. I don't know what I'm missing.

jvillafanez:
> @jvillafanez how do you get a password of user A in user encryption for the file in an anynomous request from the editor server

Is the editor server the one performing the request, or the editor server redirects the browser? If the browser is redirected, I think cookies should be still valid and we can get the information we need from the session.
Is it possible to have a workflow like (?):
1. Login in ownCloud
2. Open the file. This will open it in the editor server
3. Edit and close.
4. The editor server redirects the browser to an ownCloud's url ("/apps/richdocuments/saveFile", for example)

I expect step 4 is performed by the browser with the previous cookies. This should work at least for common case, although there are several cases we should be careful about.

mrow4a:
@jvillafanez I think we diverge from the purpose of this PR, but currently it is implemented as https://github.com/owncloud/richdocuments/blob/master/lib/Controller/DocumentController.php#L767-L773 - so this is a PublicPage and no user is logged in (and this is why we have tokens carrying user editor information https://github.com/owncloud/richdocuments/blob/master/lib/Controller/DocumentController.php#L783-L798, and this is why we have to login user to have activity app working https://github.com/owncloud/richdocuments/blob/master/lib/Controller/DocumentController.php#L1102-L1120). 

probably this is also useful explanation for @PVince81 and @sharidas

mrow4a:
@DeepDiver1975 @jvillafanez added some more clarification comments while checking for server-encryption and for default encryption module app master-key (encryption app) being enabled

micbar:
> @DeepDiver1975 @jvillafanez added some more clarification comments while checking for server-encryption and for default encryption module app master-key (encryption app) being enabled

Thank you.
User Key encryption and public pages is problematic.

Let us keep the scope of this PP

Issue #283, opened 2019-09-30
Secure view options is offered for every kind of file.


davitol:
Tested with ownCloud 10.3.0 alpha2 and  richdocuments: 2.2.0RC

Secure view options should only be offered as an option for the **mime types / file types which are associated with Collabora**. Right now it is offered for every single type of file. e.g. mp4 file shown in the screenshot below.

![Screen Shot 2019-09-30 at 14 49 16](https://user-images.githubusercontent.com/10596560/65881271-77232380-e393-11e9-89eb-af623df4ce38.png)

(cc @pmaier @piotr JFYI)

pmaier1:
AFAIK, the options should be shown for
- Folders
- ODF (.odt / .ods / .odp)
- MS Office (.docx / .xlsx / .pptx)
- PDF

@mrow4a please correct me if I missed something

mrow4a:
@pmaier1 @davitol yes, and as far as I understood: 
- for shared folders, all other formats then ODF/DOC/PDF will show exception, ODF/DOC/PDF will be opened with watermark in a safe editor
- for shared ODF/DOC/PDF, these will be opened with watermark in a safe editor

mrow4a:
@pmaier1 can you confirm that this is not a bug, but requested bahaviour, so that we can close it?

pmaier1:
> @pmaier1 can you confirm that this is not a bug, but requested bahaviour, so that we can close it?

Well, in the screenshot there is a MP4 file which can be shared with Secure View restrictions. IMO, this is not the expected behavior as only the resources in https://github.com/owncloud/richdocuments/issues/283#issuecomment-536928339 should be allowed for sharing with Secure View.

mrow4a:
@pmaier1 ahh..

Thanks for the review @davitol this is indeed a regression.

mrow4a:
fix in https://github.com/owncloud/richdocuments/pull/293

PR #293, opened 2019-10-08
[283] Make sure to register attributes only for supported mimetypes

mrow4a:
Fixes https://github.com/owncloud/richdocuments/issues/283

### DIFF
--- a/js/viewer/shareoptions.js
+++ b/js/viewer/shareoptions.js
@@ -350,8 +350,19 @@ var RichdocumentsShareOptions = {
 OC.Plugins.register('OC.Share.ShareDialogView', {
 
 	attach: function (view) {
-
+		if (_.isUndefined(view.model)) {
+			console.error("missing OC.Share.ShareItemModel");
+			return;
+		}
+		
 		RichdocumentsShareOptions.model = view.model;
+
+		// Register additional available share attributes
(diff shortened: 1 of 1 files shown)


js/viewer/shareoptions.js
```diff
@@ -350,9 +350,19 @@ var RichdocumentsShareOptions = {
 OC.Plugins.register('OC.Share.ShareDialogView', {
 
 	attach: function (view) {
-
 		RichdocumentsShareOptions.model = view.model;
 
+		if (_.isUndefined(RichdocumentsShareOptions.model)) {
+			throw 'missing OC.Share.ShareItemModel';
```
jvillafanez: Is this needed? I mean, throwing an exception will likely crash the app (I'm not sure what is the current state, but I don't remember too many "catch" blocks in javascript), which might break things due to some code not being executed.
If this is intended, I don't know the difference between throwing this exception, and letting the app crash with a "RichdocumentsShareOptions.model is undefined" just below.

I assume that the "view.model" can be undefined. Do you have a use case for this? Maybe you can include a short comment in the code.

mrow4a: Yes, probably this check is not that useful. Maybe it should just "return" instead and log warning.

PR #296, opened 2019-10-08
Do not allow to open document if viewing is disabled

mrow4a:
Do not allow to open document in Collabora if both download and viewing (e.g. with watermark) is disabled. 

- [x] strict permission checks
- [x] adjust integration tests

Cases considered:
- user receives a share with `download->false` and `view-with-watermark->true`: user can view (with watermark) the file in the editor, but cannot download to disk. 
- user receives a share with `download->false` and `view-with-watermark->false`: user cannot view the file nor download to disk (insufficient permission exception)
- user receives a share with `view-with-watermark->false`:  user cannot open editor due to viewing restriction being set to false (insufficient permission exception)
- user receives a share with `download->false`:  user can open editor with no watermark, but cannot export/download file

Screenshot at Oct 08 20-22-34

### DIFF
--- a/js/viewer/shareoptions.js
+++ b/js/viewer/shareoptions.js
@@ -10,8 +10,10 @@ var RichdocumentsShareOptions = {
 	_shareOptionsTemplate: null,
 
 	/**
-	 * Extend ShareItemModel.addShare with richdocuments attributes. This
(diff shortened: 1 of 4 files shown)


mrow4a:
@PVince81 @DeepDiver1975 can you merge?

PR #317, opened 2020-02-02
Bump version to 2.3.0 and set min-version to 10.4

timar:


micbar:
2.2.1 could do it also, you could see this as a patch release for secure view.

micbar:
👍

timar:
@micbar Unfortunately the CI won't pass with 10.4 as min-version, so I still cannot merge this.

micbar:
The CI is failing because it pulls owncloud-latest.tar.bz2 which ic owncloud 10.3.2

phil-davis:
Assigned myself so I notice this when I get to the office.
It needs to have CI adjusted to test only against the daily-master-qa tarball (which is ownCloud core 10.4.0). We know that this version of the app is not for use with the current `latest` tarball, which is 10.3.2.

phil-davis:
@micbar @timar CI is now happy.

Issue #307, opened 2019-12-22
OwnCloud menu visible when opening document via public link


T0mWz:
When an Office document is shared via a public link and will be opened in the richdocuments app, the ownCloud menu is visible.

<img width="600" alt="PublicLink" src="https://user-images.githubusercontent.com/2477797/71326411-e7f15c00-24fa-11ea-94a0-c1cc28fb366e.png">

PR #327, opened 2020-03-09
Remove header's buttons when the file is accessed through public link

jvillafanez:
Fix https://github.com/owncloud/enterprise/issues/3723 by using a custom header without any button in it. This custom header will be used only in public links, while normal accesses for registered users will use the normal header.


jvillafanez:
Code is quite complex to follow in order to add unittests, plus there are several direct calls to methods that aren't mocked at the moment (we'll likely need a refactor to fix this problem)
It doesn't seem possible to add unittests here.

adminnr1:
Is this related to https://github.com/owncloud/richdocuments/issues/307?

PR #329, opened 2020-03-12
A global config switch to (not) open in a new tab

VicDeo:
https://github.com/owncloud/richdocuments/pull/264 changed behavior to always edit documents in a new tab.
This PR adds a config switch to revert this behavior.

usage:
`php occ config:app:set richdocuments open_in_new_tab --value false`

### Doc relevant


VicDeo:
it's not possible to reach target coverage as codecov consider  2 lines in the static method to be changed

thebearon:
To be honest, I'm not convinced a configuration switch is the best way to provide an alternative to the default (and on the current default, I'm not sure it should be that editing sessions are opened on a new tab).

VicDeo:
https://github.com/owncloud/richdocuments/issues/233

lib/AppInfo/Application.php
```diff
@@ -145,10 +145,28 @@ function () {
...
+	}
+
+	public function addViewerScripts() {
+		Util::addScript('richdocuments', 'viewer/viewer');
+		Util::addStyle('richdocuments', 'viewer/odfviewer');
+	}
+
+	public function addConfigScripts($array) {
+		$appConfig = $this->getContainer()->query(AppConfig::class);
+		$array['array']['oc_appconfig']['richdocuments'] = [
```
jvillafanez: Include some comment around because it isn't clear that this will modify the array passed as parameter.

VicDeo: added as PHPDoc

jvillafanez:
This should be documented somewhere, specially taking into account that it's a setting not visible in the web UI.

mrow4a:
Yes, this was supposed to be my question, if there should be another PR for UI button

jvillafanez:
I expect most of the people not to care about this. I mean, you can easily close the tab once you've finished editing. That's why I think it's fine to have this feature a bit hidden.
If either QA or support want to have a switch in the UI, they can add a feature request to create a new PR to add the UI component.

VicDeo:
yes, I did not add it to the settings page as it's not  something that is done on a daily basis.

PR #330, opened 2020-03-17
add a background job for deleting expired tokens

karakayasemi:
No need to hold expired tokens in database, this pr adds a background job for cleaning those tokens.

Fixes #https://github.com/owncloud/enterprise/issues/3732

### DIFF
--- a/lib/BackgroundJob/CleanupExpiredWopiTokens.php
+++ b/lib/BackgroundJob/CleanupExpiredWopiTokens.php
@@ -1,0 +1,60 @@
+<?php
+/**
+ * @author Semih Serhat Karakaya <karakayasemi@itu.edu.tr>
+ *
+ * @copyright Copyright (c) 2020, ownCloud GmbH
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OCA\Richdocuments\BackgroundJob;
+
+use OC\BackgroundJob\TimedJob;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\IDBConnection;
+
+class CleanupExpiredWopiTokens extends TimedJob {
+
+	/**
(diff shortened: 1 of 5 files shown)


mrow4a:
@karakayasemi can you also, as part of this PR, fix this https://github.com/owncloud/richdocuments/blob/c2385c78e73a9ea8fe273b9e2ccc326c28e372f7/lib/Db/Wopi.php#L100 ? As you have a background job now, it is probably enough to just throw exception there.

lib/BackgroundJob/CleanupExpiredWopiTokens.php
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	/**
+	 * Makes the background job do its work
+	 *
+	 * @param array $argument unused argument
+	 */
+	public function run($argument) {
+		$now = $this->timeFactory->getTime();
+		$this->connection->executeUpdate('DELETE FROM `*PREFIX*richdocuments_wopi` WHERE `expiry` < ?',
```
mrow4a: expiry `<=`, but this is not critical.

karakayasemi:
> @karakayasemi can you also, as part of this PR, fix this
> 
> https://github.com/owncloud/richdocuments/blob/c2385c78e73a9ea8fe273b9e2ccc326c28e372f7/lib/Db/Wopi.php#L100
> ? As you have a background job now, it is probably enough to just throw exception there.

@mrow4a I think the condition was wrong in here. I corrected it and returned 404 for  this case. Please, review again. Thanks.

lib/Db/Wopi.php
```diff
@@ -96,12 +96,10 @@ public function getWopiForToken($token) {
...
 		}
 
-		//TODO: validate.
-		if ($row['expiry'] > \time()) {
+		if ($row['expiry'] <= \time()) {
 			// Expired token!
-			//http_response_code(404);
-			//$wopi->deleteBy('id', $row['id']);
-			//return false;
+			\http_response_code(404);
```
VicDeo: I don't think this side effect is needed. It's enough for `getWopiForToken` to returns `false` to controller.
And the controller should set HTTP status on false. As I see it's already done everywhere except https://github.com/owncloud/richdocuments/blob/master/lib/Controller/DocumentController.php#L879

mrow4a: @karakayasemi seems @VicDeo is right, you can remove it here and a line above, in that call just returning `false` should be sufficient as upper layer is handling response.

karakayasemi: Yes, I agree with you guys. But, the inconsistency of the method that bothers me here. The upper if block is using http_response_code also. I would remove both of them then. But, I can not be sure any possible side effect of this action.

karakayasemi: As @vicdeo stated, only one usage of the method has not any result check. I added a check for this exception place to return proper htp response code.

mrow4a: Do not worry, it is handled outside of that class

voroyam:
When can we expect the next app release containing this fix?

PR #332, opened 2020-04-20
[Tests-Only] Update phpstan to latest version

dpakach:
Related - owncloud/QA#642

Update phpstan to 0.12

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -66,7 +66,7 @@ class DocumentController extends Controller {
 	}
 
 	/**
-	 * @param \SimpleXMLElement $discovery_parsed
+	 * @param \SimpleXMLElement|null $discovery_parsed
 	 * @param string $mimetype
 	 */
 	private function getWopiSrcUrl($discovery_parsed, $mimetype) {
@@ -313,7 +313,7 @@ class DocumentController extends Controller {
 	 * - file in user folder (also shared by user/group) if fileId not null and shareToken is null
 	 * - public link (public file share or file in public folder share identified by fileId) if shareToken is not null
 	 *
-	 * @param string|null $fileId
+	 * @param string|int|null $fileId
 	 * @param string|null $shareToken
 	 * @param string $renderAs the template layout to be used
 	 * @return TemplateResponse
@@ -367,7 +367,7 @@ class DocumentController extends Controller {
 	 * - file in user folder if fileId and currently authenticated user are specified, and shareToken is null
 	 * - public link (public file share or file in public folder share identified by fileId) if shareToken is not null
 	 *
-	 * @param string|null $fileId
+	 * @param int|null $fileId
 	 * @param string|null $shareToken
 	 * @param string|null $currentUser
 	 *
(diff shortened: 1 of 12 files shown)


lib/Controller/DocumentController.php
```diff
@@ -1084,7 +1085,7 @@ private function putRelative($fileId, $owner, $editor, $suggested) {
 		$serverHost = $this->request->getServerProtocol() . '://' . $this->request->getServerHost();
 
 		// Continue editing
-		$attributes = WOPI::ATTR_CAN_VIEW | WOPI::ATTR_CAN_UPDATE | WOPI::ATTR_CAN_DOWNLOAD | WOPI::ATTR_CAN_PRINT;
+		$attributes = WOPI::ATTR_CAN_VIEW | WOPI::ATTR_CAN_UPDATE | WOPI::ATTR_CAN_PRINT;
```
dpakach, line 1088: `WOPI::ATTR_CAN_DOWNLOAD` is undefined and not so different than `WOPI::ATTR_CAN_VIEW` IMO

lib/Http/DownloadResponse.php
```diff
@@ -50,11 +53,11 @@ public function render() {
 					$this->sendNotSatisfiable($size);
```
dpakach, line 53:
```
					$this->sendNotSatisfiable($size);
```
This method is not defined and causing the phpstan to fail. not sure how to fix this

lib/Http/DownloadResponse.php
```diff
@@ -50,11 +53,11 @@ public function render() {
...
 				}
 				if ($parts[0]==='') {
-					$start = $size - $parts[1];
+					$start = $size - strlen($parts[1]);
```
dpakach, line 56: `-` operation between string and int fails in phpstan.

dpakach:
There are still calls to some undefined methods causing phpstan to fail in this repo, not sure how to fix those.

@phil-davis  please review

lib/Http/DownloadResponse.php
```diff
@@ -47,18 +50,18 @@ public function render() {
 				$parts = \explode('-', $range);
 
 				if ($parts[0]==='' && $parts[1]=='') {
-					$this->sendNotSatisfiable($size);
+					$this->sendRangeNotSatisfiable($size);
```
phil-davis, line 53: This fixes a crash that would happen if a wrong range is requested.

lib/Http/DownloadResponse.php
```diff
@@ -47,18 +50,18 @@ public function render() {
...
 					$end = $size - 1;
 				} else {
 					$start = $parts[0];
-					$end = ($parts[1]==='') ? $size - 1 : $parts[1];
+					$end = ($parts[1]==='') ? $size - 1 : (int)$parts[1];
 				}
 
 				if ($start > $end) {
-					$this->sendNotSatisfiable($size);
+					$this->sendRangeNotSatisfiable($size);
```
phil-davis, line 64: This fixes a crash that would happen if a wrong range is requested.

lib/Controller/DocumentController.php
phil-davis: I raised issue #335

PR #344, opened 2020-07-13
Avoid guzzle for discovering built-in CODE server

mrkara:
Signed-off-by: Muhammet Kara <muhammet.kara@collabora.com>

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -157,8 +157,15 @@ class DocumentController extends Controller {
 			$contact_admin = $this->l10n->t('Please contact the "%s" administrator.', [$wopiRemote]);
 
 			try {
-				$wopiClient = \OC::$server->getHTTPClientService()->newClient();
-				$discovery = $wopiClient->get($wopiDiscovery)->getBody();
+				// If we are sending query to built-in CODE server, we avoid using IClint::get() method
+				// because of an encoding issue in guzzle: https://github.com/guzzle/guzzle/issues/1758
+				if (strpos($wopiDiscovery, 'proxy.php') === false) {
+					$wopiClient = \OC::$server->getHTTPClientService()->newClient();
+					$discovery = $wopiClient->get($wopiDiscovery)->getBody();
+				}
+				else {
+					$discovery = file_get_contents($wopiDiscovery);
+				}
 			} catch (\Exception $e) {
 				$error_message = $e->getMessage();
 				if (\preg_match('/^cURL error ([0-9]*):/', $error_message, $matches)) {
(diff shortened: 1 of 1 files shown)


lib/Controller/DocumentController.php
```diff
@@ -157,8 +157,21 @@ private function getDiscovery() {
...
+				// If we are sending query to built-in CODE server, we avoid using IClint::get() method
+				// because of an encoding issue in guzzle: https://github.com/guzzle/guzzle/issues/1758
+				if (strpos($wopiDiscovery, 'proxy.php') === false) {
+					$wopiClient = \OC::$server->getHTTPClientService()->newClient();
+					$discovery = $wopiClient->get($wopiDiscovery)->getBody();
+				}
+				else {
+					$arrContextOptions=array(
+						"ssl"=>array(
+							"verify_peer"=>false,
```
DeepDiver1975, line 169: This should not be used in a live deployment - this is a potential sec risk

mrkara: > This should not be used in a live deployment - this is a potential sec risk

Thank you for the review! :)

I have adjusted the patch to enable verify_peer, and set allow_self_signed to true. That is common for test/local/small installations.

lib/Controller/DocumentController.php
```diff
@@ -157,8 +157,21 @@ private function getDiscovery() {
...
+				// because of an encoding issue in guzzle: https://github.com/guzzle/guzzle/issues/1758
+				if (strpos($wopiDiscovery, 'proxy.php') === false) {
+					$wopiClient = \OC::$server->getHTTPClientService()->newClient();
+					$discovery = $wopiClient->get($wopiDiscovery)->getBody();
+				}
+				else {
+					$arrContextOptions=array(
+						"ssl"=>array(
+							"verify_peer"=>true,
+							"allow_self_signed"=>true,
```
DeepDiver1975, line 170: just drop all these options please - we shall not allow self signed as wel. THX

lib/Controller/DocumentController.php
```diff
@@ -157,8 +157,21 @@ private function getDiscovery() {
 			$contact_admin = $this->l10n->t('Please contact the "%s" administrator.', [$wopiRemote]);
 
 			try {
-				$wopiClient = \OC::$server->getHTTPClientService()->newClient();
-				$discovery = $wopiClient->get($wopiDiscovery)->getBody();
+				// If we are sending query to built-in CODE server, we avoid using IClint::get() method
```
phil-davis, line 160:
```suggestion
				// If we are sending query to built-in CODE server, we avoid using IClient::get() method
```
Just a small typo in comment

mrkara:
Is there anything I can do to get green marks from codecov? I couldn't find anything in the tests dir, related to the getDiscovery method.

If not, or if that's not a blocker, can we merge this patch? This is prerequitsite for follow-up patches regarding the built-in server.

DeepDiver1975:
> Is there anything I can do to get green marks from codecov? I couldn't find anything in the tests dir, related to the getDiscovery method.

No - this app just lacks any kind of tests :shrug: - not much one can do here

mrkara:
I see. Thank you! :)

PR #347, opened 2020-07-24
Switch to new license Manager

micbar:
## Description
- Switch to new license manager
- Raise minimum dependency to 10.5

## To do

- [x] Fix tests 

@pmaier1

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -13,6 +13,7 @@
 
 namespace OCA\Richdocuments;
 
+use OC\License\LicenseManager;
 use OCP\App\IAppManager;
 use \OCP\IConfig;
 
@@ -28,10 +29,12 @@ class AppConfig {
 
 	private $config;
 	private $appManager;
+	private $licenseManager;
 
-	public function __construct(IConfig $config, IAppManager $appManager) {
+	public function __construct(IConfig $config, IAppManager $appManager, LicenseManager $licenseManager) {
 		$this->config = $config;
 		$this->appManager = $appManager;
+		$this->licenseManager = $licenseManager;
 	}
 
 	/**
@@ -86,7 +89,7 @@ class AppConfig {
 	 * @return bool
 	 */
 	public function enterpriseFeaturesEnabled() {
-		if (!$this->appManager->isEnabledForUser('enterprise_key') && !\getenv('CI')) {
+		if (!$this->licenseManager->checkLicenseFor($this->appName)) {
 			return false;
 		}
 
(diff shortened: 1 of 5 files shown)


phil-davis:
In the description: "Run acceptance tests on php7.2 an php7.4"

That is already happening - codestyle, phpunit and acceptance tests are all running on PHP 7.2, 7,3, 7.4
So that point is not happening in this PR.

micbar:
@phil-davis I needed to remove the tests on `latest`

We need to re enable after the core release

phil-davis:
> @phil-davis I needed to remove the tests on `latest`
> 
> We need to re enable after the core release

:+1: this had to be done on all repos that have min-version 10.5

Some time after 10.5.0 release becomes "latest" core, there will be a dozen repos to have tests against "latest" core enabled again.

micbar:
@phil-davis Created #348

lib/AppConfig.php
```diff
@@ -86,7 +89,7 @@ public function setUserValue($userId, $key, $value) {
 	 * @return bool
 	 */
 	public function enterpriseFeaturesEnabled() {
-		if (!$this->appManager->isEnabledForUser('enterprise_key') && !\getenv('CI')) {
+		if (!$this->licenseManager->checkLicenseFor($this->appName)) {
```
mrow4a, line 92: `&& !\getenv('CI')` still required

micbar: Not anymore. We have the grace period now which works also in CI

phil-davis: For reference: https://github.com/owncloud/core/blob/master/tests/acceptance/run.sh#L774
Starts tests with the grace period popup disabled.

mrow4a:
Weird I cannot use the app with new License Manager:
```
➜  octest git:(master) ✗ ls lib/private/License   
BasicLicense.php   ILicense.php       LicenseFetcher.php LicenseManager.php MessageService.php
```

I get this

```
Exception: {\"Exception\":\"OCP\\\\AppFramework\\\\QueryException\",\"Message\":\"Could not resolve uid! Class uid does not exist or error resolving constructor arguments: Could not resolve LicenseManager! Class LicenseManager does not exist\
```

I will try to figure out what is wrong with your PR

mrow4a:
@micbar @pmaier1 fixed richdocuments, but requires merging https://github.com/owncloud/core/pull/37728

Tested with secure view
Zrzut ekranu 2020-07-26 o 10 26 17

micbar:
Core PR is merged. The new daily master tarballs are build this night.

So tomorrow the CI should pass.

phil-davis:
Rebased so CI includes the test change in #349

PR #354, opened 2020-08-10
Hotfix for checking license for a specific feature in richdocuments

jvillafanez:
Fixes https://github.com/owncloud/enterprise/issues/4143

### Use cases:

* Upgrade ownCloud CE to 10.5.0 with richdocuments having this patch
  The app will keep working as CE. No grace period will be started and no license is needed
* Upgrade ownCloud EE to 10.5.0 with richdocuments having this patch
  **admin MUST explicitly set the app as EE**, (see the "expected consequences" section):
  ```
  occ config:app:set richdocuments isEEActive --value yes
  ```
  This could be done before upgrading
* Enabling richdocument's EE features
  Run the following command. The grace period will be started and a license will be asked for (normal grace period workflow)
  ```
  occ config:app:set richdocuments isEEActive --value yes
  ```

### Expected consequences of this patch:
* For CE: none. The app will behave normally without any problem
* For EE:
  * They should run the `occ config:app:set richdocuments isEEActive --value yes` before upgrading to 10.5.0 to ensure there won't be problems caused by a temporary removal of the secure view feature.
  * In case they forget to run the command:
    * It seems the secure view will keep working if it was active before (to be verified)
    * Settings page will show a "Enable Secure View (requires Enterprise edition)". This might be fine. Running the above command should fix this
    * You can uncheck the "Enable Secure View" checkbox, but you won't be able to recheck it again if the page is reloaded. You'll need to run the command above.

### Upgrade path
It's expected that 10.5.1 (or the version following 10.5.0) will address this problem properly.
Once this new OC version is out, it's expected that richdocuments will undo these fixes automatically via migration, as well as adjust the code to work properly with 10.5.1

### DIFF
--- a/lib/AppConfig.php
+++ b/lib/AppConfig.php
@@ -89,7 +89,19 @@ class AppConfig {
 	 * @return bool
 	 */
 	public function enterpriseFeaturesEnabled() {
-		if (!$this->licenseManager->checkLicenseFor($this->appName)) {
+		$isEnterprise = $this->config->getAppValue($this->appName, 'isEEActive', null);
+		if ($isEnterprise === null) {
+			// app not marked as EE -> return false without doing anything more
+			return false;
+		}
+
+		$isEnterprise = \filter_var($isEnterprise, FILTER_VALIDATE_BOOLEAN);
+		if (!$isEnterprise || !$this->licenseManager->checkLicenseFor($this->appName)) {
+			// if not marked as enterprise, no need to check the license:
+			// explicitly mark the app as non-enterprise and remove the secure view.
+			// if marked as enterprise, we need to verify the license
+			$this->config->setAppValue($this->appName, 'isEEActive', 'no');
+			$this->config->setAppValue($this->appName, 'secure_view_option', 'false');
 			return false;
 		}
 
--- a/tests/acceptance/features/webUISecureView/main.feature
+++ b/tests/acceptance/features/webUISecureView/main.feature
@@ -5,26 +5,30 @@ Feature: Secure View
   So that users can share files with very restricted access to documents
 
   Scenario: Admin enables the secure view through webUI
-    Given the administrator has browsed to the admin additional settings page
+    Given the administrator has added config key "isEEActive" with value "yes" in app "richdocuments"
+    And the administrator has browsed to the admin additional settings page
     When the administrator enables secure view using the webUI
     Then the config key "secure_view_option" of app "richdocuments" should have value "true"
 
   Scenario: Admin enables the secure view, enabled print permissions through webUI
-    Given the administrator has browsed to the admin additional settings page
+    Given the administrator has added config key "isEEActive" with value "yes" in app "richdocuments"
+    And the administrator has browsed to the admin additional settings page
     When the administrator enables secure view using the webUI
     And the administrator enables print permission in secure view using the webUI
     Then the config key "secure_view_option" of app "richdocuments" should have value "true"
     And the config key "secure_view_can_print_default" of app "richdocuments" should have value "true"
 
   Scenario: Admin enables the secure view, disables secure-view permissions through webUI
-    Given the administrator has browsed to the admin additional settings page
+    Given the administrator has added config key "isEEActive" with value "yes" in app "richdocuments"
+    And the administrator has browsed to the admin additional settings page
     When the administrator enables secure view using the webUI
     And the administrator disables watermark permission in secure view using the webUI
     Then the config key "secure_view_option" of app "richdocuments" should have value "true"
     And the config key "secure_view_has_watermark_default" of app "richdocuments" should have value "false"
 
   Scenario: Admin enables the secure view using occ command, disables print default and enables it again through webUI
-    Given the administrator has added config key "secure_view_option" with value "true" in app "richdocuments"
+    Given the administrator has added config key "isEEActive" with value "yes" in app "richdocuments"
+    And the administrator has added config key "secure_view_option" with value "true" in app "richdocuments"
     And the administrator has added config key "secure_view_can_print_default" with value "false" in app "richdocuments"
     And the administrator has browsed to the admin additional settings page
     When the administrator enables print permission in secure view using the webUI
@@ -32,7 +36,8 @@ Feature: Secure View
     And the config key "secure_view_can_print_default" of app "richdocuments" should have value "true"
 
   Scenario: Admin enables the secure view using occ command, disables secure-view default and enables it again through webUI
-    Given the administrator has added config key "secure_view_option" with value "true" in app "richdocuments"
+    Given the administrator has added config key "isEEActive" with value "yes" in app "richdocuments"
+    And the administrator has added config key "secure_view_option" with value "true" in app "richdocuments"
     And the administrator has added config key "secure_view_has_watermark_default" with value "false" in app "richdocuments"
     And the administrator has browsed to the admin additional settings page
     When the administrator enables watermark permission in secure view using the webUI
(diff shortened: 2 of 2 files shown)


jnweiger:
The changes look okay to me. The situation with `occ config:app:set richdocuments isEEActive --value yes` is not nice, But the impact of missing this step seems low.

This occ command should be mentioned in the CHANGELOG. but this app does not have a CHANGELOG. Please add one, if possible.

lib/AppConfig.php
```diff
@@ -89,7 +89,19 @@ public function setUserValue($userId, $key, $value) {
...
+			// app not marked as EE -> return false without doing anything more
+			return false;
+		}
+
+		$isEnterprise = \filter_var($isEnterprise, FILTER_VALIDATE_BOOLEAN);
+		if (!$isEnterprise || !$this->licenseManager->checkLicenseFor($this->appName)) {
+			// if not marked as enterprise, no need to check the license:
+			// explicitly mark the app as non-enterprise and remove the secure view.
+			// if marked as enterprise, we need to verify the license
+			$this->config->setAppValue($this->appName, 'isEEActive', 'no');
```
mrow4a, line 103: lets align with true/false ?

lib/AppConfig.php
```diff
@@ -89,7 +89,19 @@ public function setUserValue($userId, $key, $value) {
...
+			$this->config->setAppValue($this->appName, 'secure_view_option', 'false');
 			return false;
 		}
```
mrow4a, line 107: what happens in case of true? Wont we need to set app values in this case too ?

lib/AppConfig.php
jvillafanez: My main worry is with the type-checking. The expectation is that the admin runs the `occ config:app:set richdocuments isEEActive --value yes` command. I guess this will be evaluated and saved as string. Even if they set `true` or `false` as value, it will be a string. This might be confusing if we get a boolean as default because we'll need to deal with true/false as boolean and also with true/false as string. Note that `"false"` == `true`

By using yes/no, it's more clear that they'll always be string.
In addition, DB value is also more clear (it will show yes/no instead of 1 and empty string) and also is aligned with other similar keys.

lib/AppConfig.php
jvillafanez: If there is no "isEEActive" key, the code won't reach this point. The app will be considered as CE.
The admin will need to explicitly set the key via the `config:app:set` command.
If he set the value as "yes", we'll check the license and revert the value if the license is wrong. We'll also revert the "secure_view_option" because the app is considered now as CE due to the wrong license.
If he set the value as "yes" and the license is valid, we don't need to change anything. The admin must have changed the "isEEActive" value manually, so we don't need to change it, and we also don't want to modify the other flags such as "secure_view_option" (the app could be EE but secure_view might not be active)

lib/AppConfig.php
mrow4a: Yes you are right, we should change this for other setting too

lib/AppConfig.php
mrow4a: I still am blury. Can you write below flow in points: - enabling richdocuments when having enterprise license, - enabling richdocuments not having the license

jvillafanez:
> The changes look okay to me. The situation with occ config:app:set richdocuments isEEActive --value yes is not nice, But the impact of missing this step seems low.

I'd rather confirm it by the QA team. The expectation is that CE won't have any impact, and for EE we can email the clients if needed

> This occ command should be mentioned in the CHANGELOG. but this app does not have a CHANGELOG. Please add one, if possible.

I guess this will need additional setup if we want to align it with the rest of the changelogs. We could include it in the readme, but I don't think it will have enough visibility.
The good point is that, as said, this is expected to affect only to EE, so we could send an email.

In theory, the plan is to run the command only once. Upgrade from this version to the one for 10.5.1 will have a migration to cleanup the setting automatically. In case an upgrade from a previous version to 10.5.1, they shouldn't need to run anything.
My point is that this should be known, but I don't think is too critical to make big changes around this.

lib/AppConfig.php
jvillafanez: * No "isEEActive" key => do nothing. Only community features **should** enabled (upgrade is a special case)
* "isEEActive" = "yes":
  * if valid license => enable EE features
  * if invalid license => set "isEEActive" = "no" and disable EE features (secure_view will be false)
* "isEEActive" = "no" => set "isEEActive" = "no" and disable EE features

For the upgrade case, QA should verify that it works as EE if there is a valid license. This scenario can fall into
* "isEEActive" = "yes" + valid license (as instructed, no problems expected)
* no "isEEActive" key (need to ensure nothing wrong happens)

In this last scenario, I've noticed that there is a "license is required" message for the secure_view, but it still shows as active if it was active before, as well as the rest of the related options. In case it wasn't active it's possible that the checkbox remains disabled. This is fixed by setting the "isEEActive" = "yes"

mrow4a: It is not fully my question. I wamt to know what *system-admin* has to do given just TAR with everything, starting from scratch.

jvillafanez: For a fresh installation, the admin won't have to do anything if he just want to use CE features. If he want to use EE features, he'll need to provide a valid license and set the "isEEActive" = "yes".

In case of an upgrade, if he wants CE features, he won't need to do anything. If the current version has EE features enabled, he should run the occ command before the upgrade

mrow4a: @pmaier is this acceptable? To me sounds fishy.

jvillafanez:
Checked upgrade having a valid license and with secure_view in-use:
* OC 10.4.1 to 10.5.0
* enterprise_key app 0.2.1 to master
* richdocuments 2.3.0 to this branch

1. Setup OC 10.4.1 with richdocuments 2.3.0 (2.4.0 isn't compatible)
2. Setup collabora
3. Set a valid OC license in ownCloud (you'll need the enterprise_key app, tested with version 0.2.1)
4. Configure the connection with collabora. Check that you can open files with it from ownCloud.
5. Enable secure_view in the richdocuments' settings (settings -> admin-> additional)
6. Upload / create a file
7. Share it with User1 / Group1 (public links doesn't have secure_view option (?)). Ensure the secure_view option is checked
8. Login with that User1 and open the shared file. The configured watermark should appear for that user
9. Upgrade the server (**the admin forgets to run the occ command** - which could happen):
    * OC 10.4.1 => 10.5.0
    * enterprise_key 0.2.1 => master
    * richdocuments 2.3.0 => this branch

Consequences:
* User1 can still access to the shared file with the watermark being shown (same behaviour as before)
* Admin can still change the secure_view checkbox in the file, as well as the "can print" one.
* Secure_view is still marked as checked in the settings page. Related options can still be changed at will. You can check / uncheck the secure view until you reload the page (see below)
* Disabling the option and reloading the page will uncheck the secure_view permanently.
  * You need to reload the page to trigger this.
  * Running the occ command will fix this.
* "Enable Secure View (requires Enterprise edition)" appears in the settings page.
  * This is just a misleading message. The feature works if it was active before
  * Running the occ command will fix this.

Screenshot showing the state after the upgrade:
Screenshot from 2020-08-13 11-37-58


Even if the admin forgets to run the command, I think this is acceptable, just a couple of misleading things that can be fixed easily.

pmaier1:
> 7. Share it with User1 / Group1 (public links doesn't have secure_view option (?)

Secure View is only available for user/group sharing. Please check that reshare privilege escalation is prevented (no public links on a secure view shares / no resharing with higher permission; no resharing at all? I don't remember the latest state)

jvillafanez:
If the secure_view is checked in the file, the share permission is unchecked (as well as edit and change), so the receiving user can't share at all.

jnweiger:
Thanks for the tests!
That confirms that an enterprise server can be upgraded to 10.5.0. Having done that, the admin would also want to upgrade to richdocuments-2.4.0 -- is that also safe?

jvillafanez:
2.4.0 could be considered (wrongly) as an enterprise app: as long as there is a valid license, all features are available.
Problems come when there isn't a valid license. The license check isn't in a good place for an "all-features-are-enterprise" app, so the behaviour is a bit difficult to predict, but the app will likely be disabled at some point.
It's safe as long as the license is valid. If the license is invalid, we'd have to test what happens.

PR #358, opened 2020-09-10
Very minimal document templates for Collabora Online

timar:
These files contain no data at all. This guarantees that when the
user opens a new document in the app, the language of paragraph,
page size, cell date format, currency, etc. will be according to
the current locale.

Signed-off-by: Andras Timar <andras.timar@collabora.com>


phil-davis:
@micbar please assign the appropriate person to review this.

jvillafanez (approved):

PR #361, opened 2020-10-24
Fix if statement reported by phpstan

phil-davis:
The latest release of phpstan reports:
```
php -d zend.enable_gc=0 vendor-bin/phpstan/vendor/bin/phpstan analyse --memory-limit=4G --configuration=./phpstan.neon --no-progress --level=5 appinfo lib
⚠️  You're using a deprecated config option bootstrap. ⚠️️

This option has been replaced with bootstrapFiles which accepts a list of files
to execute before the analysis.

 ------ -------------------------------------------------------------------- 
  Line   lib/Controller/DocumentController.php                               
 ------ -------------------------------------------------------------------- 
  78     Comparison operation ">" between int<1, max> and 0 is always true.  
 ------ -------------------------------------------------------------------- 

                                                                                                                        
 [ERROR] Found 1 error                                                                          
```

1) Adjust the `if` statement to have a more strict comparison
2) update the phpstan setting to use `bootstrapFiles`

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -75,7 +75,7 @@ class DocumentController extends Controller {
 		}
 
 		$result = $discovery_parsed->xpath(\sprintf('/wopi-discovery/net-zone/app[@name=\'%s\']/action', $mimetype));
-		if ($result && \count($result) > 0) {
+		if (($result !== false) && (\count($result) > 0)) {
 			return [
 				'urlsrc' => (string)$result[0]['urlsrc'],
 				'action' => (string)$result[0]['name']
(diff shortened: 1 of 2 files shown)


phil-davis:
There are very little unit tests. What to do about codecov fail?

mrow4a:
@phil-davis I would need to test that if statement (or at least go deep in the code), php is very bad in such comparisons and things can break.

phil-davis:
The "problem" was that the xpath call returns either boolean `false` or an array. If `$result` is an empty array, then that is considered `false` by a check like `if ($result)`. So the only way for `$result` to be true (the first item in the original `if` test) is if it is a non-empty array. So the 2nd check `\count($result) > 0` makes no difference - it will always be true if we get past the 1st check.

IMO what I did is a clear way to handle the possible return values of `xpath`

mrow4a (approved):

PR #362, opened 2020-10-25
refresh the token lifetime in every access with that token

karakayasemi:
If the expiry is closer than configured time, refresh the token lifetime in next access with that token

Fixes https://github.com/owncloud/enterprise/issues/4237

### DIFF
--- a/lib/Db/Wopi.php
+++ b/lib/Db/Wopi.php
@@ -20,6 +20,8 @@ namespace OCA\Richdocuments\Db;
 class Wopi extends \OCA\Richdocuments\Db {
 	// Tokens expire after this many seconds (not defined by WOPI specs).
 	const TOKEN_LIFETIME_SECONDS = 36000;
+	// If the expiry is closer than this time, it will be refreshed.
+	const TOKEN_REFRESH_THRESHOLD_SECONDS = 3600;
 
 	const ATTR_CAN_VIEW = 0;
 	const ATTR_CAN_UPDATE = 1;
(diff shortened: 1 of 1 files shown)


mrow4a (changes requested):
Please do not refresh on each access, check expiration time from row, and renew only when approaches expiry or e.g. after X hours of last renewal.

karakayasemi:
@mrow4a thank you for review. I changed the logic and now it will refresh token only if the expiry is closer than `TOKEN_REFRESH_THRESHOLD_SECONDS`.

karakayasemi:
@mrow4a the pr is ready to review again, please help by reviewing it one more time. Thanks.

mrow4a (commented):
generally on how token refresh usually works, and if we do this renewing I would do it properly.
- on each access, check if token needs renewal
- if can be renewed (e.g. token lifetime of x days was not exceeded), do renewal. Kerberos snd OAuth tokens e.g. can be renewed by 1 day up to 7 days. 1 token should not live infinitely for security reasons
- next, check for expiry of the token

In your current code, you first check expiry, and then attempt renewal. Let me know your thoughts.

mrow4a:
@karakayasemi I had a look again. Main point is not to have a single token to have possibly infinite lifetime (due to refreshes). What would be the point of expiry then, if you could renew the same token infinitelly ? In fact, probably user as part of refresh should receive new token to be used (or be somehow forced to relogin?). 

what do you think about this all? just some suggestions/feedback

karakayasemi:
@mrow4a your points are reasonable, I read the wopi documentation about this topic again and I realized that we just need to return `access_token_ttl` value for wopi clients. In this way, wopi client will handle rest of the things like warning user when token close to expire and refreshing tokens. Currently, we do not return any `access_token_ttl` and because of that, wopi clients think that token lifetime is infinite. 

I am working on a PR for `access_token_ttl`. I will update here when I finish it. Thanks.

karakayasemi:
Closing this one. Let's continue with #364 .

PR #364, opened 2020-11-04
pass access token ttl value to wopi client

karakayasemi:
Wopi protocol suggests returning access_token_ttl value to wopi client. In this way, wopi client can decide what to do when token about to expire. 
For example, collabora warns user when 15 minutes before expire and shows below message to user:
resim

If user still doesn't refresh the session, it warns again periodically in every 2 minutes.

Fixes owncloud/enterprise#4237

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -265,11 +265,14 @@ var documentsMain = {
 			    "&permission=readonly";
 
 			// access_token - must be passed via a form post
-			var access_token = encodeURIComponent(documentsMain.token);
+			var access_token = encodeURIComponent(documentsMain.accessToken);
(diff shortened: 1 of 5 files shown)


mrow4a (approved):
exactly how I would expect it to work. Great work! Unit tests not straightforward I belive?

karakayasemi:
There is no dependency injection in php classes. Also, touched classes have too many `\OC::$server`, static method usage etc. JS side has not single unit test. Yes, nearly imposible to test without big refactoring. Seems like test can be written only to newly added classes in this repository. Maybe @micbar can help to merge as it is.

karakayasemi:
@phil-davis Could you help for merging this pull request?

PR #368, opened 2020-11-19
Fix custom app folder document create problem

karakayasemi:
If the richdocuments app is installed in the custom folder and Use OOXML by default for new files option is enabled, richdocuments app could not create any document from template. This pr fixes explained problem.

Fixes https://github.com/owncloud/enterprise/issues/4281


mrow4a (approved):

karakayasemi:
Fixed the broken unit test. Again I attempted to write unit test for the changes, but a big refactoring needed to able to write test.

PR #371, opened 2020-11-23
make sure accessing a file triggers auditable read operation

mrow4a:
Context: when collaborating on the file, only 1st read operation is being performed (for the user opening the document). Next readers were opening directly the editor.

This PR adds read operation when aquiring token for a user.

Tested:
- make sure audit log is triggered 

fixes https://github.com/owncloud/enterprise/issues/4290

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -18,6 +18,7 @@ use \OCP\AppFramework\Controller;
 use \OCP\Constants;
 use OCP\Files\File;
 use OCP\IGroupManager;
+use OCP\Files\NotPermittedException;
 use \OCP\IRequest;
 use \OCP\IConfig;
 use \OCP\IL10N;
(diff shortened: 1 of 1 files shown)


lib/Controller/DocumentController.php
```diff
@@ -805,12 +805,21 @@ public function wopiCheckFileInfo($documentId) {
 			return new JSONResponse([], Http::STATUS_NOT_FOUND);
 		}
 
+		// make sure file can be read when checkinf file info
```
phil-davis, line 808:
```suggestion
		// make sure file can be read when checking file info
```

DeepDiver1975 (approved):

DeepDiver1975 (approved):
Squash please.
:+1: for the rest. Nice job!

lib/Controller/DocumentController.php
```diff
@@ -805,12 +805,21 @@ public function wopiCheckFileInfo($documentId) {
...
 		$file = $this->getFileHandle($fileId, $res['owner'], $res['editor']);
 		if (!$file) {
 			$this->logger->warning('wopiCheckFileInfo(): Could not retrieve file', ['app' => $this->appName]);
 			return new JSONResponse([], Http::STATUS_NOT_FOUND);
 		}
 
+		// trigger read operation while checking file info for user
+		// after acquiring the token
+		$fileHandle = $file->fopen('rb');
+		if (!$fileHandle) {
```
phil-davis, line 818: https://drone.owncloud.com/owncloud/richdocuments/1133/4/4
"818    Negated boolean expression is always false."
`phpstan` thinks that the result of `$file->fopen('rb');` can never be false.

PR #382, opened 2021-02-25
Federated shares support

eszkadev:


mmattel:
Added reviewers so it will not get overseen

mrow4a:
Cherry-picked commits from here and extanded basic handling from here with "federated handshake" -> https://github.com/owncloud/richdocuments/pull/498

Issue #385, opened 2021-03-02
Cannot open docs shared from Local Storage


theheyon:
Hey on there!

System: Ubuntu 18.04.4 LTS + ownCloud 10.6.0 + CollaboraOnline 2.4.1
CODE:  LOOLWSD 6.4.6 / LOKit Collabora Office 6.4-23

When sharing folders with Public Links - everything goes fine until you try to share folder from **Local storage**. Then something goes wrong - getting error when you click on file:
```Collabora Online: Cannot open document. Backends provided no user object for```

Sharing any folder or document from any user folder seems to be ok.
Maybe, some trouble with missing folder owner or something like that?

Thanks for any direction to fix that =)

**Comment by theheyon** [comment]
Answering to myself =)
Figured out that same issue were fixed in NextCloud app:
https://github.com/nextcloud/richdocuments/commit/ee09fb07482e288c3245fb0948220ad6cedf3c32

So i patched _/lib/Storage.php_ a bit and its seems works...

Part of original Storage.php:
```
// Retrieve user folder for the file to be able to get relative path
$owner = $document->getOwner()->getUID();
$root = \OC::$server->getRootFolder()->getUserFolder($owner);

$ret = [];
$ret['owner'] = $document->getOwner()->getUID();
$ret['permissions'] = $share->getPermissions();
$ret['updateable'] = $document->isUpdateable();
$ret['mimetype'] = $document->getMimeType();
$ret['path'] = $root->getRelativePath($document->getPath());
$ret['name'] = $document->getName();
$ret['fileid'] = $document->getId();
$ret['instanceid'] = \OC::$server->getConfig()->getSystemValue('instanceid');
$ret['version'] = '0'; // latest
$ret['sessionid'] = '0'; // default shared session

return $ret;
```

Patched:
```
// Retrieve user folder for the file to be able to get relative path

$owner = $share->getShareOwner();
if (is_null($owner)) {
    $owner = $document->getOwner()->getUID();
}
$root = \OC::$server->getRootFolder()->getUserFolder($owner);

$ret = [];
$ret['owner'] = $owner;
$ret['permissions'] = $share->getPermissions();
$ret['updateable'] = $document->isUpdateable();
$ret['mimetype'] = $document->getMimeType();
$ret['path'] = $root->getRelativePath($document->getPath());
$ret['name'] = $document->getName();
$ret['fileid'] = $document->getId();
[...]

PR #386, opened 2021-03-05
Fix Public Links shared from Local Storage

theheyon:
https://github.com/owncloud/richdocuments/issues/385


mrow4a (approved):
makes total sense! With external storages $document->getOwner() will return null for some reason, would need to investigate. As this code is executed only for shares, it is safest to take this info from share directly.

tested and it works properly

mrow4a:
@jnweiger merged fyi

PR #400, opened 2021-05-21
Allow setting default open action with secure-view enforced session for supported documents

JanAckermann:
Fixes: https://github.com/owncloud/enterprise/issues/4578 , https://github.com/owncloud/enterprise/issues/4566

### Description
With this change, it will be possible to activate the Setting "Open documents in Secure View with watermark by default".
Due to this change, the file owner can view/edit/export/print a file in secure view mode (With watermark) without sharing the file.

image

image

image

image

### DIFF
--- a/js/viewer/viewer.js
+++ b/js/viewer/viewer.js
@@ -45,56 +45,39 @@ var odfViewer = {
 		return (odfViewer.supportedMimes.indexOf(mimetype) !== -1);
 	},
 
-	register : function(){
-		var i,
-		    mime;
+	register : function() {
+		for (var i = 0; i < odfViewer.supportedMimes.length; ++i) {
+			var mime = odfViewer.supportedMimes[i];
 
-		for (i = 0; i < odfViewer.supportedMimes.length; ++i) {
-			mime = odfViewer.supportedMimes[i];
-			OCA.Files.fileActions.registerAction(
-				{
-					name: "Richdocuments",
-					actionHandler: odfViewer.onEdit,
+			if(!$("#isPublic").val() &&
+					OC.appConfig.richdocuments &&
+					OC.appConfig.richdocuments.secureViewAllowed === true &&
+					OC.appConfig.richdocuments.secureViewOpenActionDefault === true) {
+				OCA.Files.fileActions.registerAction({
+					name: 'RichdocumentsSecureView',
+					actionHandler: odfViewer.onOpenWithSecureView,
+					displayName: t('richdocuments', 'Open in Collabora with Secure View'),
+					iconClass: 'icon-lock-closed',
+					permissions: OC.PERMISSION_READ,
+					mime: mime
+				});
+				OCA.Files.fileActions.setDefault(mime, 'RichdocumentsSecureView');
+			} else {
+				OCA.Files.fileActions.registerAction({
+					name: 'Richdocuments',
+					actionHandler: odfViewer.onOpen,
 					displayName: t('richdocuments', 'Open in Collabora'),
 					iconClass: 'icon-rename',
-					permissions: 	OC.PERMISSION_UPDATE | OC.PERMISSION_READ,
+					permissions: OC.PERMISSION_READ,
 					mime: mime
-				}
-			);
-
-			if(OC.appConfig.richdocuments &&
-				OC.appConfig.richdocuments.secureViewAllowed === true &&
-				OC.appConfig.richdocuments.defaultShareAttributes &&
-				OC.appConfig.richdocuments.defaultShareAttributes.secureViewHasWatermark === true
-			){
-				OCA.Files.fileActions.registerAction(
-					{
-						name: "RichdocumentsSecureView",
-						actionHandler: odfViewer.onEditSecureView,
-						displayName: t('richdocuments', 'Open in Collabora with Secure View'),
-						iconClass: 'icon-lock-closed',
-						permissions: OC.PERMISSION_UPDATE | OC.PERMISSION_READ,
-						mime: mime
-					}
-				);
+				});
+				OCA.Files.fileActions.setDefault(mime, 'Richdocuments');
 			}
-
-			OCA.Files.fileActions.setDefault(mime, 'Richdocuments');
 		}
 
 	},
 
-	dispatch : function(filename){
-		if (this.isSupportedMimeType(OCA.Files.fileActions.getCurrentMimeType())
-			&& OCA.Files.fileActions.getCurrentPermissions() & OC.PERMISSION_UPDATE
-		){
-			odfViewer.onEdit(filename);
-		} else {
-			odfViewer.onView(filename);
-		}
-	},
-
-	onEdit : function(fileName, context){
+	onOpen : function(fileName, context){
 		var fileId = context.$file.attr('data-id');
 		var fileDir = context.dir;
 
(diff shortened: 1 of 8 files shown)


JanAckermann:
@hodyroff 
> File click will now show 2 actions (Open in Collabora & Open in Collabora Secure-View)

We can overcome this with a small change in the core. After the improvement, clicking the file will opens directly in Collabora if only Collabora is assigned to the mime type. Open in Secure view will be available in the 3 dot menu.

> Disable the option if file is a share and shared with secure-view-option ?
This is not doable without a big refactor! There will be always a 'Open in Collabora' and 'Open in Collabora with Secure View' which will always do the same if the share is secure view protected

mrow4a:
@JanAckermann great work, just needed few details (registerAction permission) and handling for wopi document attributes! Screens
Zrzut ekranu 2021-05-24 o 16 34 49

Zrzut ekranu 2021-05-24 o 16 35 12
Zrzut ekranu 2021-05-24 o 16 35 25
cc @hodyroff

JanAckermann:
@mrow4a thank you, I am about to developing a core change that the appdrawer don't show the Open in secure View, @hodyroff and customer needs to decide what fits best

mrow4a:
@JanAckermann I think that we should not overengineer this.. I strongly vote against changes in core for this change, result will be the same.

JanAckermann:
@mrow4a this will be a common change and the product can profit.

mrow4a:
@JanAckermann what about delivering the fix to customer and any further improvemrnts put under the decision matrix? Lets improve continously and where strongly needed/requested.

mrow4a:
btw Jan, not sure if you noticed but requirememt for „ Disable the option if file is a share and shared with secure-view-option „ is also covered. You missed fact that you can filter options with permission filters. E.g read-only docs will have standard „Open in Collabora” that if file is shared and secure view is set will open with it. Files with edit righs will have both standard option „Open in Collabora” that will work on whatever permissions are set for file , and „Open in Collabora with enforced Secure View” that will force Secure View on file with edit permissions.

JanAckermann:
@mrow4a 
I am ok with it. #400 is a bugfix, everything else an improvemt, do we need tests or can we call it ready?

mrow4a:
@JanAckermann this PR is not a bugfix. It is a feature that was working due to a bug.

JanAckermann:
>btw Jan, not sure if you noticed but requirememt for „ Disable the option if file is a share and shared with secure-view-option „ is also covered. You missed fact that you can filter options with permission filters. E.g read-only docs will have standard „Open in Collabora” that if file is shared and secure view is set will open with it. Files with edit righs will have both standard option „Open in Collabora” that will work on whatever permissions are set for file , and „Open in Collabora with enforced Secure View” that will force Secure View on file with edit permissions.

Tested the following:
(Secure-View enabled, Watermark set)
1. User a creates a odt file
2. User a shares with user b, only secure-view option enabled
3. User b clicks on the shared file
4. App switcher is opening and asks us if we want to open with collabora or collabora with secure view

**Additionally**
Now the user b can print the file event if print/export is not enabled, while clicking "Open in Collabora with Secure-View", I tested this before and it was working as expected, I don't know if the issue comes from the new docker image or your changes 

image

image

mrow4a:
@JanAckermann can you hard reload and clean cache? I think your JS is outdated. Check JS code in console.

JanAckermann:
@mrow4a Private window + cleaned + disabled cache
results in the same problem

mrow4a:
@JanAckermann can you show me in the console JS code for this https://github.com/owncloud/richdocuments/pull/400/files#diff-982c3683c43a8fb4c83489c33d580c6d51f9a304c76da6cf5380f351b18f9180R70-R77 ? richdocuments/js/viewer/viewer.js

JanAckermann:
image

@mrow4a JS code seems to be up2date

JanAckermann:
@mrow4a I have no idea, just shared via ui now magic in here, also found out that I can open secure-view files with files_texteditor

mrow4a:
@JanAckermann I cannot reproduce this, and it makes no sense. Permission assigments in OC are far outside of this change, and 50% of integration tests in core would fail. Something is wrong with your dev environment.

JanAckermann:
@mrow4a I will reset it and try it again ;)

JanAckermann:
@mrow4a checked out core latest master, removed config.php removed data folder, same issue!

mrow4a:
@JanAckermann this change is only related to rendering a windows when permission is set, and setting custom flag server side. Problem you describe I have not seen and in core there are integration tests for it. 

Test it outside of richdocuments (disable the app), try editing and if issue persists create issue/escalate.

JanAckermann:
@mrow4a reproduced without richdocuments, @JammingBen can you confirm?

JammingBen:
> @mrow4a reproduced without richdocuments, @JammingBen can you confirm (Share a file without any permissions, and rename will rendered anyways)?

Yes, we also have two issues about this behavior: https://github.com/owncloud/enterprise/issues/4582 and https://github.com/owncloud/core/issues/30325

mrow4a:
@JanAckermann @JammingBen interesting. I tested the problem sharing "folders" with secure-view. But I did not try sharing specific files. Maybe there is a problem in core? Thoughs?

JanAckermann:
@mrow4a there is going on a lot of stuff with the permissions, If I share a file without any persmissions, the fileinfo permission is 11 (apps/files_sharing/js/share.js:44)  which is a combination of PERMISSION_READ & PERMISSION_UPDATE & PERMISSION_DELETE

can you check with a specific file to confirm it is an issue?

mrow4a:
@JanAckermann I can confirm what you found for shared single files, that there is some bug in core that somehow assigns PERMISSION_UPDATE to file that has been shared without PERMISSION_UPDATE . Folders are not affected.

mrow4a:
@JanAckermann @hodyroff 

New admin setting `Open documents in Secure View with watermark by default`:
Zrzut ekranu 2021-05-25 o 23 38 37

For secure-view only:
Zrzut ekranu 2021-05-25 o 23 37 13

For editors:
Zrzut ekranu 2021-05-25 o 23 19 39

JanAckermann:
@mrow4a I like the solution! 
The only downside is that you as a file owner, don't have the ability to open with or without secure-view, but this is related to the core permission issue.

js/viewer/viewer.js
```diff
@@ -45,38 +45,39 @@ var odfViewer = {
...
-			mime = odfViewer.supportedMimes[i];
-			OCA.Files.fileActions.registerAction(
-				{
-					name: "Richdocuments",
-					actionHandler: odfViewer.onEdit,
+	register : function() {
+		for (var i = 0; i < odfViewer.supportedMimes.length; ++i) {
+			var mime = odfViewer.supportedMimes[i];
+
+			if(!$("#isPublic").val() &&
```
jvillafanez, line 52: Either the naming is terrible wrong, or the condition is wrong. I assume that, in a public link, this condition should be true, and in a public link I guess we want to prioritize the secure view (if the rest of the conditions apply)

jvillafanez:
It might be a good idea to have a list of things that have been tested, as confirmation. A quick sum up of the changes could also help (I'm not fully sure how changing the setting will affect the app)

Code looks good so far, but I'm not sure if it behaves as expected or not.

JanAckermann:
@jvillafanez AFAIK secure-view with public links is not supported yet and this behaviour will not be take in account in this PR

JammingBen (approved):
LGTM code-wise 👍

mmattel:
Docs relevant, please file a docs issue.
@EParzefall fyi

PR #404, opened 2021-06-07
Enable comments on pdfs 

JanAckermann:
Before this PR ist was not possible to add comments on PDF files with rich documents/Collabora, this works now.

Fixes: - https://github.com/owncloud/enterprise/issues/4606

Before:
image

After:
image

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -426,7 +426,7 @@ class DocumentController extends Controller {
 
 		// Update permissions
 		$permissions = $doc['permissions'];
-		if (!($doc['action'] === 'edit')) {
+		if (!($doc['action'] === 'edit') && !($doc['action'] === 'view_comment')) {
 			$permissions = $permissions & ~\OCP\Constants::PERMISSION_UPDATE;
 		}
 


JanAckermann:
https://github.com/CollaboraOnline/online/commit/ca00470722ec7a7a7b190ca36a1f71da06ed6847

lib/Controller/DocumentController.php
```diff
@@ -426,7 +426,7 @@ private function handleDocIndex($fileId, $shareToken, $currentUser) {
 
 		// Update permissions
 		$permissions = $doc['permissions'];
-		if (!($doc['action'] === 'edit')) {
+		if (!($doc['action'] === 'edit') && !($doc['action'] === 'view_comment')) {
```
mrow4a, line 429: it is **&&** or **||** ? hm

JammingBen: `&&` is correct 🙂 

The action for PDFs is `view_comment`, and we want to include the update permission in this case. With `||` we would run into the if-block which would remove this permission.

JanAckermann: As @JammingBen pointed out, the line:
`$permissions & ~\OCP\Constants::PERMISSION_UPDATE;`
removes the PERMISSION_UPDATE from the pemissions var

jvillafanez (approved):

PR #409, opened 2021-06-15
Prevent documents with tabs in filenames / or any other invalid chars from being created 

JanAckermann:
Fixes: https://github.com/owncloud/enterprise/issues/4627

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -17,6 +17,7 @@ use OCP\App\IAppManager;
 use \OCP\AppFramework\Controller;
 use \OCP\Constants;
 use OCP\Files\File;
+use OCP\Files\InvalidPathException;
 use OCP\IGroupManager;
 use OCP\Files\NotPermittedException;
 use \OCP\IRequest;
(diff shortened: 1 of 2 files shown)


lib/Controller/DocumentController.php
```diff
@@ -1131,6 +1142,16 @@ private function putRelative($fileId, $owner, $editor, $suggested) {
 			$root->newFolder(\dirname($path));
 		}
 
+		try {
+			$view = new View();
+			$view->verifyPath($path, $suggested);
+		} catch (\Exception $e) {
```
jvillafanez, line 1148: Better catch only the `InvalidPathException`, which is the one we're interested in. Other exceptions might require different handling

lib/Controller/DocumentController.php
```diff
@@ -1131,6 +1142,16 @@ private function putRelative($fileId, $owner, $editor, $suggested) {
 			$root->newFolder(\dirname($path));
 		}
 
+		try {
+			$view = new View();
```
jvillafanez, line 1146: any alternative on creating the view instance here?
In addition, creating the view for the root folder might be confusing if the path used in the `verifyPath` function isn't relative to the root folder. Someone could try a `$view->getFileInfo($path)` later, that could fail because the path refers to the user's root folder, not to the root folder.
Furthermore, it makes more sense that the `verifyPath` checks the information based on the specific view. I mean, a path might be valid for the root folder, but not for other folders. I don't think we have use cases like this, but it feels more natural.

mrow4a (approved):
perfectly done!

PR #437, opened 2021-12-15
Make upload work again

JanAckermann:
Upload in rich documents office overview stopped working, fixed with this change. 

Fixes:
- https://github.com/owncloud/enterprise/issues/4869#event-5770193716

### DIFF
--- a/templates/documents.php
+++ b/templates/documents.php
@@ -48,25 +48,11 @@ script('files', 'jquery.fileupload');
 				<label><?php p($l->t('New Presentation')) ?></label>
 			</a>
 			<div id="upload" title="<?php p($l->t('Upload (max. %s)', [$_['uploadMaxHumanFilesize']])) ?>">
-				<form data-upload-id="1"
-					  id="data-upload-form"
-					  class="file_upload_form"
-					  action="<?php print_unescaped(link_to('files', 'ajax/upload.php')); ?>"
-					  method="post"
-					  enctype="multipart/form-data"
-					  target="file_upload_target_1">
-					<?php if ($_['uploadMaxFilesize'] >= 0):?>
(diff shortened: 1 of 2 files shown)


templates/documents.php
```diff
@@ -48,25 +48,10 @@
...
-					  id="data-upload-form"
-					  class="file_upload_form"
-					  action="<?php print_unescaped(link_to('files', 'ajax/upload.php')); ?>"
-					  method="post"
-					  enctype="multipart/form-data"
-					  target="file_upload_target_1">
-					<?php if ($_['uploadMaxFilesize'] >= 0):?>
-					<input type="hidden" name="MAX_FILE_SIZE" id="max_upload"
-						   value="<?php p($_['uploadMaxFilesize']) ?>" />
-					<?php endif;?>
```
JanAckermann, line 61: seems like this was never set before, only accessible while opening a file

templates/documents.php
```diff
@@ -48,25 +48,10 @@
...
-					<!-- Send the requesttoken, this is needed for older IE versions
-						 because they don't send the CSRF token via HTTP header in this case -->
-					<input type="hidden" name="requesttoken" value="<?php p($_['requesttoken']) ?>" id="requesttoken" />
```
JanAckermann, line 64: Need to check with IE11

mrow4a (approved):

templates/documents.php
JanAckermann: Works with IE11 not needed

cdamken:
I tested in damken-cloud
now it works! Thanks @JanAckermann @mrow4a !!

jnweiger:
Confirmed fixed in 2.7.0 RC2 with both classic and new Web UI.

Issue #460, opened 2022-03-17
phpstan errors


phil-davis:
https://github.com/phpstan/phpstan/releases/tag/1.4.10 was recently released and is automatically used in CI. It does more/better code analysis.

The latest phpstan is reporting:
https://drone.owncloud.com/owncloud/richdocuments/1842/3/4
```
Installing dependencies from lock file (including require-dev)
Package operations: 1 install, 0 updates, 0 removals
  - Downloading phpstan/phpstan (1.4.10)
  - Installing phpstan/phpstan (1.4.10): Extracting archive
Generating autoload files
1 package you are using is looking for funding.
Use the `composer fund` command to find out more!
php -d zend.enable_gc=0 vendor-bin/phpstan/vendor/bin/phpstan analyse --memory-limit=4G --configuration=./phpstan.neon --no-progress --level=5 appinfo lib
 ------ --------------------------------------- 
  Line   lib/Controller/DocumentController.php  
 ------ --------------------------------------- 
  1245   Right side of && is always true.       
 ------ --------------------------------------- 

 ------ --------------------------------- 
  Line   lib/Storage.php                  
 ------ --------------------------------- 
  167    Left side of && is always true.  
 ------ --------------------------------- 

 [ERROR] Found 2 errors                                                         

make: *** [test-php-phpstan] Error 1
Makefile:168: recipe for target 'test-php-phpstan' failed
```

phil-davis:
PR #462 fixes this.

PR #462, opened 2022-03-17
Adjust 'if' conditionals that were reported by phpstan

phil-davis:
Fixes issue #460 and gets CI green again.

### DIFF
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -1242,7 +1242,7 @@ class DocumentController extends Controller {
 	 * @return null|\OCP\Files\File
 	 */
 	private function getFileHandle($fileId, $owner, $editor) {
-		if ($editor && $editor != '') {
+		if ($editor) {
 			$user = $this->userManager->get($editor);
 			if (!$user) {
 				$this->logger->warning('wopiPutFile(): No such user', ['app' => $this->appName]);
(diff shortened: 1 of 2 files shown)


lib/Controller/DocumentController.php
```diff
@@ -1242,7 +1242,7 @@ private function putRelative($fileId, $owner, $editor, $suggested) {
 	 * @return null|\OCP\Files\File
 	 */
 	private function getFileHandle($fileId, $owner, $editor) {
-		if ($editor && $editor != '') {
+		if ($editor) {
```
phil-davis, line 1245: If `$editor` is the empty string, then `$editor` will evaluate to `false` in the first condition, so the 2nd condition adds no value.

lib/Storage.php
```diff
@@ -164,7 +164,9 @@ public function getDocumentByShareToken($token, $fileId = null) {
 			if ($node->getType() == FileInfo::TYPE_FILE) {
 				/** @var \OCP\Files\Node|null $document */
 				$document = $node;
-			} elseif ($node->getType() == FileInfo::TYPE_FOLDER && $fileId !== null) {
+			} elseif ($fileId !== null) {
```
phil-davis, line 167: At this point we know that the resource type is not TYPE_FILE. `getType` can only return 2 values, so we know that `$node` must be of type `TYPE_FOLDER`. Therefore that check is removed.

jvillafanez (approved):

PR #475, opened 2022-08-15
WIP: add support for link share preview

mrow4a:
WARNING: do not merge, yet this is in Proof of Concept state only

Screenshot 2022-08-15 at 21 51 56
Screenshot 2022-08-15 at 21 52 38
Screenshot 2022-08-15 at 21 52 53
Screenshot 2022-08-15 at 21 59 03

### DIFF
--- a/js/viewer/shareoptions.js
+++ b/js/viewer/shareoptions.js
@@ -393,3 +393,67 @@ OC.Plugins.register('OC.Share.ShareDialogView', {
 		});
 	}
 });
+
+
+OC.Plugins.register('OCA.Share.ShareDialogLinkShareView', {
+
+	attach: function (view) {
+		if (_.isUndefined(view.model)) {
+			console.error("missing OC.Share.ShareItemModel");
+			return;
+		}
+
+		// NOTE: folder is not supported at the moment
+		var mimetype = view.itemModel.getFileInfo().getMimeType();
+		if (!odfViewer.isSupportedMimeType(mimetype)) {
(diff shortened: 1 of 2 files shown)


JanAckermann (approved):
To me the wording 'Preview' is very misleading, I would vote for 'View only' at least in the UI.
Other than that, LGTM

NannaBarz:
@mrow4a what should we do here in the Serverteam board?

mrow4a:
@NannaBarz decide whether we want to have this feature or not.. ref. https://github.com/owncloud/enterprise/issues/5277

jnweiger:
@hodyroff this is not in the richdcuments 4.0.0 release ticket. Should it be added?

mrow4a:
@jnweiger this here was PoC. Also requires Core change.

pako81:
@mrow4a would be a lot of effort to finalize this + https://github.com/owncloud/core/pull/40289 based on the original approach? Would be very nice to have this in core 10.13 + richdocuments 4.0.0.

pako81:
@mrow4a @hodyroff @jnweiger should we consider moving this forward now?

hodyroff:
No, good POC. But the use case for secure view/no download are very secret documents. Why would a user show those through public links, rather then accounts which can be audited and tracked? We continue the work when this comes up as a requirement again.

hodyroff:
Yes, not valid anymore. This is a public repo!

PR #483, opened 2022-12-07
[tests-only] [full-ci] Run phan with php 7.3

phil-davis:
`phan` will report errors if any of the code is not compatible with PHP 7.3.

This app still supports PHP 7.3 when used with oC10 core 10.11 and earlier.

### DIFF
--- a/lib/Db/Storage.php
+++ b/lib/Db/Storage.php
@@ -12,10 +12,6 @@
 
 namespace OCA\Richdocuments\Db;
 
-/**
- * @method string loadRecentDocumentsForMimes()
- */
-
 class Storage extends \OCA\Richdocuments\Db {
 	public const appName = 'richdocuments';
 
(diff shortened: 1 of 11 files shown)


lib/Db/Storage.php
```diff
@@ -12,10 +12,6 @@
 
 namespace OCA\Richdocuments\Db;
 
-/**
- * @method string loadRecentDocumentsForMimes()
- */
-
```
phil-davis, line 18: This method is already a normal method in this class. phan complains about it being declared here in PHPdoc - that is not necessary. So I deleted this PHPdoc comment.

vendor-bin/phan/composer.json
```diff
@@ -1,10 +1,5 @@
 {
-  "config": {
-    "platform": {
-      "php": "7.1"
-    }
-  },
```
phil-davis, line 6: This was committed in the original setup of the phan infrastructure a long time ago.
We don't want that any more. We want phan to use whatever PHP version is the default at run-time (so it will run appropriately in each CI pipeline).

saw-jan (approved):

PR #486, opened 2023-02-28
E5515 feature/wopi locks

mrow4a:
- [x] Separate DocumentController (routes for collabora for owncloud app) and WopiController (routes for WOPI protocol implementation)
- [x] In WopiController change `\OC::$server->X` to class variables so unit testing is possible
- [x] Separate getting file handle into dedicated class FileService, and change `\OC::$server->X` to class variables so unit testing is possible
- [x] Rename `Storage` -> `DocumentService`
- [x] WopiController CheckFileInfo now returns `'SupportsGetLock' => false, 'SupportsLocks' => false,` as preparation to implement locking workflow as in https://learn.microsoft.com/en-us/microsoft-365/cloud-storage-partner-program/rest/files/lock?source=recommendations
- [ ] Implement Lock, RefreshLock, Unlock methods in WopiController
- [ ] add unit tests
- [ ] make sure existing acceptance test pass


lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+			'app' => $this->appName,
+			'fileId' => $fileId,
+			'token' => $token ]);
+
+		$row = new Db\Wopi();
+		$row->loadBy('token', $token);
+
+		//TODO: Support X-WOPIMaxExpectedSize header.
+		$res = $row->getWopiForToken($token);
+		if ($res == false) {
```
jvillafanez, line 277: `===` or comment

lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+			'app' => $this->appName,
+			'fileId' => $fileId,
+			'token' => $token
+		]);
+
+		$row = new Db\Wopi();
+		$row->loadBy('token', $token);
+
+		$res = $row->getWopiForToken($token);
+		if ($res == false) {
```
jvillafanez, line 314: `===` or comment

lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+				'app' => $this->appName,
+				'headerTime' => $wopiHeaderTime,
+				'storageTime' => Helper::toISO8601($file->getMtime())
+			]);
+			// Tell WOPI client about this conflict.
+			return new JSONResponse(['LOOLStatusCode' => self::LOOL_STATUS_DOC_CHANGED], Http::STATUS_CONFLICT);
+		}
+
+		// Read the contents of the file from the POST body and store.
+		$content = \fopen('php://input', 'r');
```
jvillafanez, line 356: Is it possible to use the `OCP\IRequest` instead?

lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+		$this->logger->debug('PutFileRelative: file {fileId}, token {token}.', [
+			'app' => $this->appName,
+			'fileId' => $fileId,
+			'token' => $token]);
+
+		$row = new Db\Wopi();
+		$row->loadBy('token', $token);
+
+		$res = $row->getWopiForToken($token);
+		if ($res == false) {
```
jvillafanez, line 394: `===` or comment

lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+			return new JSONResponse([], Http::STATUS_FORBIDDEN);
+		}
+
+		// get owner and editor uid's
+		$owner = $res['owner'];
+		$editor = $res['editor'];
+
+		// Retrieve suggested target
+		$suggested = $this->request->getHeader('X-WOPI-SuggestedTarget');
+		$suggested = \iconv('utf-7', 'utf-8', $suggested);
```
jvillafanez, line 411: I think we'll need a comment here. Unless it's strictly needed, I'd rather not touch the encoding.
Even better if you can provide a link to an official source to prove that utf-7 is used.

lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+		$this->logger->debug('Lock: file {fileId}, wopiLock {wopiLock}, token {token}.', [
+			'app' => $this->appName,
+			'fileId' => $fileId,
+			'wopiLock' => $wopiLock,
+			'token' => $token]);
+
+		$row = new Db\Wopi();
+		$row->loadBy('token', $token);
+		$res = $row->getWopiForToken($token);
+		if ($res == false) {
```
jvillafanez, line 515: `===` or comment

lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+		 * @phpstan-ignore-next-line
+		*/
+		'@phan-var IPersistentLockingStorage $storage';
+		$locks = $storage->getLocks($file->getInternalPath(), false);
+
+		// handle non-existing lock
+
+		if (empty($locks)) {
+			// set new lock
+			if (isset($editor) && $editor != '') {
```
jvillafanez, line 544: `!==` or comment

lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+		$this->logger->debug('Unlock: file {fileId}, wopiLock {wopiLock}, token {token}.', [
+			'app' => $this->appName,
+			'fileId' => $fileId,
+			'wopiLock' => $wopiLock,
+			'token' => $token]);
+
+		$row = new Db\Wopi();
+		$row->loadBy('token', $token);
+		$res = $row->getWopiForToken($token);
+		if ($res == false) {
```
jvillafanez, line 619: `===` or comment

lib/Controller/WopiController.php
```diff
@@ -0,0 +1,873 @@
...
+		/**
+		 * @var IPersistentLockingStorage $storage
+		 * @phpstan-ignore-next-line
+		*/
+		'@phan-var IPersistentLockingStorage $storage';
+		$storage->unlockNodePersistent($file->getInternalPath(), [
+			'token' => $wopiLockOld,
+		]);
+
+		if (isset($editor) && $editor != '') {
```
jvillafanez, line 845: `!==` or comment

lib/FileService.php
```diff
@@ -0,0 +1,156 @@
...
+namespace OCA\Richdocuments;
+
+use OC\User\Session;
+use OCP\Files\IRootFolder;
+use OCP\Files\File;
+use OCP\ILogger;
+use OCP\IUserManager;
+use OCA\Richdocuments\AppConfig;
+use Symfony\Component\EventDispatcher\GenericEvent;
+use Symfony\Component\EventDispatcher\EventDispatcher;
```
jvillafanez, line 32: It might be better to use the `EventDispatcherInterface`

lib/FileService.php
```diff
@@ -0,0 +1,156 @@
...
+			if (!$this->appConfig->encryptionEnabled()) {
+				// Set session for a user
+				$this->userSession->setUser($user);
+			} elseif ($this->appConfig->masterEncryptionEnabled()) {
+				// With master encryption, decryption is based on master key (no user password needed)
+				// make sure audit/activity is triggered for editor session
+				$this->userSession->setUser($user);
+
+				// emit login event to allow decryption of files via master key
+				$afterEvent = new GenericEvent(null, ['loginType' => 'password', 'user' => $user, 'uid' => $userUID, 'password' => '']);
```
jvillafanez, line 116: I guess we'll need to think about this at some point because a login event shouldn't be triggered here.
Probably no short term action to do for now, but we shouldn't figure out something to do this properly.

lib/AppInfo/Application.php
```diff
@@ -51,12 +52,27 @@ private function registerServices() {
...
+			/** @var IContainer $c */
+			return new WopiController(
+				$c->query('AppName'),
+				$c->query('Request'),
+				$c->query('CoreConfig'),
+				$c->query('AppConfig'),
+				$c->query('L10N'),
+				$c->query('Logger'),
+				$c->query('FileService'),
+				$c->query('ServerContainer')->getRootFolder(),
```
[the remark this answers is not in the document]
mrow4a: refactor the whole class for this, makes sense

lib/Controller/WopiController.php
mrow4a: this was copied over, would prefer not to refactor this part as I already refactored a lot

lib/FileService.php
mrow4a: Yes, we would need to do this in Core. I think this was discussed longer time ago, but I think this would need a rework of the encryption stuff in core to do this correctly, which is a bit risky.

mrow4a: adjusted

jvillafanez:
I expected something like:
```
$container->registerService('WopiController', function (SimpleContainer $c) {
			$server = $c->getServer();
			return new WopiController(
				$c->query('AppName'),
.....
				$c->query('FileService'),
				$server->getRootFolder(),
				$server->getURLGenerator(),
				$server->getUserManager()
			);
		});
```
I'm not sure if there is any drawback with any of the approaches. Maybe in your code, the server will be fixed (it won't change) when the constructor is run, so if the server changes later it could be a problem (the reference will still point to the old server, not the new one), but this shouldn't be a realistic scenario because there is no reason to change the server reference.
Could you check other apps in order to homogenize the behavior? If there isn't any common ground, you can keep this.

In terms of memory usage, I guess it could be a couple of additional references that need to stay in memory, but the object itself can't be freed, so this shouldn't be a problem.

jvillafanez (approved):
Code looks fine, just minor things.

PR #493, opened 2023-05-25
make sure to retrieve correct supershare based on current dir context

mrow4a:
fixes https://github.com/owncloud/enterprise/issues/5729

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -223,7 +223,7 @@ var documentsMain = {
 		if (dir)
 			ocurl = OC.generateUrl('apps/richdocuments/index?fileId={fileId}&dir={dir}', {fileId: fileId, dir: dir});
 		else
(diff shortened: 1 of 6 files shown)


mrow4a:
@pako81 this is rather straighforward change , if you could have a look.

pako81:
Code-wise looks good to me.

pako81 (approved):

pako81:
Additionally, tested manually scenario described in https://github.com/owncloud/enterprise/issues/5729#issuecomment-1556702074 with this code in place and it works as expected: now share permissions are correctly evaluated 👍

pako81:
@mrow4a is this dependent on a specific core version? @jnweiger we may need a new richdocuments app release (see https://github.com/owncloud/enterprise/issues/5729 for Customer's urgency).

mrow4a:
works with current core version.

jnweiger:
@mrow4a related compatibility question: will it explode when loaded in 10.11.0 or will it just work as "broken" as before?

PR #494, opened 2023-05-29
Refactor API and most critical parts of the code

mrow4a:
this is done as preparation for https://github.com/owncloud/richdocuments/pull/382. logic does not change, pure refactor.

changes:
- [x] notify that major version (4.0.0?) needed to be raised for the app @jnweiger 
- [x] deprecate `/wopi/extapp` endpoint as not used and maintained, cleanup related code
- [x] cleanup routes.php to reflect purpose of the APIs
- [x] annotate legacy methods that are hard to refactor 
- [x] refactor and modernize DocumentsController, including split of logic to dedicated classes
- [x] add unit tests

remark: for review use Split Review Mode


mrow4a:
@jvillafanez if you could have a review here. I basically extracted some logic into dedicated classes so I can actually add tests for them, and/or simply code. Some depreciations of not used logic has also been done.

lib/Controller/DocumentController.php
```diff
@@ -321,48 +154,95 @@ private function domainOnly($url) {
...
+		$renderAs = 'user';
+
+		// Handle general response
+		$wopiRemote = $this->discoveryService->getWopiUrl();
+		if (($parts = \parse_url($wopiRemote)) && isset($parts['scheme'], $parts['host'])) {
+			$webSocketProtocol = "ws://";
+			if ($parts['scheme'] == "https") {
+				$webSocketProtocol = "wss://";
+			}
+			$webSocket = \sprintf(
```
jvillafanez, line 180: what about something like
```
$webSocket = "{$wsProtocol}://{$parts['host']}";
if (isset($parts['port'])) {
  $webSocket .= ":{$parts['port']}";
}
```

lib/Controller/DocumentController.php
```diff
@@ -871,50 +732,47 @@ private function getWopiInfoForPublicLink(array $docInfo) : array {
...
+			$documents[$key]['hasPreview'] = $this->previewManager->isMimeSupported($document['mimetype']);
+			$documents[$key]['urlsrc'] = $wopiSrcUrl['urlsrc'];
+			$documents[$key]['action'] = $wopiSrcUrl['action'];
+			$documents[$key]['locale'] = $locale;
 		}
 
-		return new JSONResponse([], Http::STATUS_UNAUTHORIZED);
-	}
+		\usort($documents, function ($a, $b) {
+			return @$b['mtime']-@$a['mtime'];
```
jvillafanez, line 771: we should include a comment here. I know we're suppressing the error message because we know the key might be missing, but taking into account this isn't common, I don't know what the value will be.
It's better to include some explanation about what's happening behind the scenes and what is the expected results. For example, I don't know if the documents without mtime will be first or last.

lib/DiscoveryService.php
```diff
@@ -0,0 +1,192 @@
...
+				if (\strpos($wopiDiscovery, 'proxy.php') === false) {
+					$wopiClient = $this->httpClient->newClient();
+					$discovery = $wopiClient->get($wopiDiscovery)->getBody();
+				} else {
+					$discovery = \file_get_contents($wopiDiscovery);
+				}
+			} catch (\Exception $e) {
+				$error_message = $e->getMessage();
+
+				if (\preg_match('/^cURL error ([0-9]*):/', $error_message, $matches)) {
```
jvillafanez, line 137: Any reason we don't send the error message directly to the log file? I'd rather not parse messages

lib/DiscoveryService.php
```diff
@@ -0,0 +1,192 @@
...
+	 *
+	 * @return SimpleXMLElement|null returns discovery or null in case of discovery cannot be retrieved
+	 */
+	private function getDiscovery() : ?SimpleXMLElement {
+		if ($this->appConfig->testUserSessionEnabled()) {
+			$wopiRemote = $this->appConfig->getAppValue('test_wopi_url');
+			$discoveryKey = 'discovery.xml_test';
+		} else {
+			$wopiRemote = $this->appConfig->getAppValue('wopi_url');
+			$discoveryKey = 'discovery.xml';
```
jvillafanez, line 112: `discoveryKey` seems a bad name. I though it was some kind of secret or password, not just a filename.

lib/DiscoveryService.php
```diff
@@ -0,0 +1,192 @@
...
+				return null;
+			}
+
+			if (!$discovery) {
+				$this->logger->error('Collabora Online: Unable to read discovery.xml from {wopiRemote}', ['app' => 'richdocuments', 'wopiRemote' => $wopiRemote]);
+				return null;
+			}
+
+			$this->logger->debug('Storing {discoveryKey} to the cache.', ['app' => 'richdocuments', 'discoveryKey' => $discoveryKey]);
+			$this->cache->set($discoveryKey, $discovery, 3600);
```
jvillafanez, line 175: just to confirm, we don't need any kind of encryption nor encoding for the stored value, do we?

lib/DocumentService.php
```diff
@@ -90,16 +90,57 @@ public function __construct(
...
+				continue;
+			}
+
+			$document = [
+				'fileid' => $fileId,
+				'path' => $path,
+				'name' => $fileName,
+				'mimetype' => $mimeType,
+				'mtime' => $mtime
+				];
```
jvillafanez, line 134: wrong indentation

jvillafanez:
In addition, check the vulnerability shown in sonarcloud. We should solve it somehow

lib/AppInfo/Application.php
```diff
@@ -110,6 +113,19 @@ private function registerServices() {
 			);
 		});
 
+		$container->registerService('DiscoveryService', function (SimpleContainer $c) use ($server) {
+			$config = $c->query('AppConfig');
+			$logger = $server->getLogger();
+			$cacheFactory = $c->query('ICacheFactory');
```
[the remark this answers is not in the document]
mrow4a: got rid of that Application.php legacy way of initializing now. thanks!

lib/AppConfig.php
```diff
@@ -166,4 +167,34 @@ public function secureViewHasWatermarkDefaultEnabled() {
...
+		$user = \OC::$server->getUserSession()->getUser();
+		if ($user === null) {
+			return false;
+		}
+
+		$uid = $user->getUID();
+		$testgroups = \array_filter(\explode('|', $this->getAppValue('test_server_groups')));
+		foreach ($testgroups as $testgroup) {
+			$test = \OC::$server->getGroupManager()->get($testgroup);
+			if ($test !== null && \sizeof($test->searchUsers($uid)) > 0) {
```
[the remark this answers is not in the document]
mrow4a: not sure if I understand. In fact this function is years old, and I am actually not sure if used at all.

lib/Controller/DocumentController.php
```diff
@@ -321,48 +154,95 @@ private function domainOnly($url) {
...
+		$fileId = \is_numeric($fileId) ? (int)$fileId : null;
+		
 		// Normal editing and user/group share editing
 		// Parameter $dir is not used during indexing, but might be used by Document Server
-		return $this->handleIndex($fileId, $dir, null, 'user');
+		$renderAs = 'user';
+
+		// Handle general response
+		$wopiRemote = $this->discoveryService->getWopiUrl();
+		if (($parts = \parse_url($wopiRemote)) && isset($parts['scheme'], $parts['host'])) {
```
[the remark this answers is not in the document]
mrow4a: done

lib/Controller/DocumentController.php
mrow4a: `{$parts['host']}` is not so good.

lib/Controller/DocumentController.php
mrow4a: very old code, lets not touch this. LEGACY tag is in the header.

lib/DiscoveryService.php
mrow4a: Hard to judge, but if someone added it, lets keep this.

lib/DocumentService.php
mrow4a: done

mrow4a:
@jvillafanez adjusted

lib/AppConfig.php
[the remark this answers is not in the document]
jvillafanez: The `testUserSessionEnabled` says that it checks if the session user is a tester, but this line searches for users. There could be more than one user, the search could consider additional search terms... there is no guarantee that the search will be an exact match of an unique uid. Double-check with the user_ldap app.

lib/Controller/DocumentController.php
```diff
@@ -233,25 +236,32 @@ public function index($fileId, $dir) {
 	 * @PublicPage
 	 */
 	public function public($shareToken, $fileId) {
-		// If type of fileId is a string, then it
-		// doesn't work for shared documents, lets cast to int everytime
-		$fileId = \is_numeric($fileId) ? (int)$fileId : null;
+		if (\is_string($shareToken) && \strlen($shareToken) > 0 && \is_numeric($fileId)) {
```
jvillafanez, line 239: checks should match with the ones in the "index" method. Not sure if these are the good ones or not.

mrow4a: @jvillafanez why so? these methods are different
```
public function public($shareToken, $fileId)
public function index($fileId, $dir)
```

jvillafanez: my bad. I didn't read the parameters correctly.

lib/AppConfig.php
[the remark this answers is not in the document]
mrow4a: Lets leave it for a moment, I have no even idea how to test this. Seems a bit legacy functionality.

jvillafanez (approved):

PR #495, opened 2023-06-05
CI changes and update coding-standard to v4

phil-davis:


phpstan.neon
```diff
@@ -6,9 +6,15 @@ parameters:
     - %currentWorkingDirectory%/appinfo/Migrations/*.php
     - %currentWorkingDirectory%/appinfo/routes.php
   ignoreErrors:
+    -
+      message: '#^Right side of || is always false.$#'
+      path: lib/Controller/DocumentController.php
+      count: 1
```
phil-davis, line 12: This is a new thing that phpstan is finding.
IMO the code can be easily and safely fixed, but actually the reported problem does no harm, so I won't touch real code in this PR.

saw-jan (approved):

PR #496, opened 2023-06-09
Migrate to oC Web v7

kulmann:
## Preface
ownCloud Web version 7.0.0 has been released just recently. It does not support Vue2 (EOL end of 2023) anymore as it moved on to Vue3. Furthermore there have been some improvements in the extension system for editor apps.

## Description
Making `richdocuments` compatible with ownCloud Web v7.

**Warning**: there was a fix necessary, so this only works with ownCloud Web v7.0.2 or higher.

## How to test as a dev
With the ownCloud Web dev stack this PR can be used for testing and development as follows:

1. clone the `richdocuments` repo and then check out this branch
2. check out the `stable-7.0` branch of ownCloud Web
3. adjust some files in oC Web:
- file `dev/docker/oc10.entrypoint.sh`, add `occ app:enable richdocuments` at the end of the file
- file `dev/docker/oc10.web.config.json`, add richdocuments as external app:
  ```json
  "external_apps": [
      {
        "id": "richdocuments",
        "path": "http://host.docker.internal:8080/index.php/apps/richdocuments/js/richdocuments.js"
      }
   ]
  ```
- file `docker-compose.yml`, add your local clone of this repo as volume to the oc10 container: `- /your/local/path/to/richdocuments:/mnt/data/apps/richdocuments`
4. `docker-compose up -d oc10`
5. `docker run -t -d -p 9980:9980 -e "extra_params=--o:ssl.enable=false" -e "username=admin" -e "password=admin" --name collabora --cap-add MKNOD collabora/code:6.4.8.6`
6. add `http://host.docker.internal:9980` as collabora url in the oc10 admin settings (last menu item `Additional`, then text field `Collabora Online server`)
7. switch to `New Design` in the ownCloud app switcher
8. try all the collabora features.


src/editor.vue
```diff
@@ -0,0 +1,213 @@
...
+  useAccessToken,
+  useAppDefaults,
+  useConfigurationManager,
+  useStore
+} from '@ownclouders/web-pkg'
+import { defineComponent, ref } from 'vue'
+
+export default defineComponent({
+  setup() {
+    const store = useStore()
```
kulmann, line 36: Results in error `TypeError: Cannot read properties of null (reading 'proxy')` because `getCurrentInstance()` (= used deep down inside the `useStore` composable) is null... any idea how to work around that? 🤔

dschmidt: https://github.com/owncloud/richdocuments/pull/496/commits/35b1822351416529ad45eefdecd71b17ea4ee7d7 🙈

kulmann: Impressively awful 😅 thank you!

lib/Controller/WebAssetController.php
```diff
@@ -62,7 +61,7 @@ public function get(): Response {
 
 		try {
 			return new DataDisplayResponse(\file_get_contents($filePath), Http::STATUS_OK, [
-				'Content-Type' => $this->getMimeType($filePath),
+				'Content-Type' => 'text/javascript',
```
kulmann, line 64: @xoxys cc

kulmann:
@dschmidt could I get a code review for this? testing it after merge with oc10 and oc web v7.0.2 and kicking of a release for the richdocuments app is then probably something for @jnweiger and team 🙏

lib/Controller/WebAssetController.php
dschmidt: looks little fishy tbh

dschmidt (approved):
I don't like that we have to check in built artifacts but cleaning that up is probably way beyond out of scope of this PR.

I'm happy with the javascript build setup part otherwise.

The php change is hard to judge for me.

lib/Controller/WebAssetController.php
dschmidt: discussed in zoom, I'm fine with it for this super simple case

kulmann: and for the record: in [web](https://github.com/owncloud/web/blob/acfc3165b56fadd16db85e7aaefaea8cb1d0e98d/packages/web-integration-oc10/lib/Controller/FilesController.php#L61) we inject a mimetype detector that comes along from within oc10. I decided to not use that because I don't expect that we'll ever serve another file than this single javascript file.

PR #497, opened 2023-06-10
handle federated share mount to display error and further refactor

mrow4a:
- [x] make sure not to add or remove any other feature (this should be just refactor)
- [x] further refactor needed for future work in https://github.com/owncloud/richdocuments/pull/382
- [x] add missing handling for federated share that should display error
- [x] add extensive tests

Screenshot 2023-06-09 at 16 16 27

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -164,8 +164,6 @@ var documentsMain = {
 	isEditorMode : false,
 	isViewerMode: false,
 	fileName: null,
-	canShare : false,
-	canEdit: false,
 	loadError : false,
 	loadErrorMessage : '',
 	loadErrorHint : '',
@@ -487,7 +485,7 @@ var documentsMain = {
 			    "&lang=" + OC.getLocale().replace('_', '-') +
 			    "&closebutton=1" +
 			    "&revisionhistory=1";
-			if (!documentsMain.canEdit || action === "view") {
+			if (action === "view") {
 				urlsrc += "&permission=readonly";
 			}
 
@@ -706,8 +704,6 @@ var documentsMain = {
 
 		$(documentsMain.toolbar).appendTo('#header');
 
-		documentsMain.canShare = typeof OC.Share !== 'undefined' && rd_permissions & OC.PERMISSION_SHARE;
-
 		// fade out file list and show the document
 		$('#content-wrapper').fadeOut('fast').promise().done(function() {
 			documentsMain.fileId = rd_fileId;
@@ -715,7 +711,6 @@ var documentsMain = {
 			documentsMain.version = rd_version;
 			documentsMain.sessionId = rd_sessionId;
 			documentsMain.fileName = rd_title;
-			documentsMain.canEdit = Boolean(rd_permissions & OC.PERMISSION_UPDATE);
 
 			documentsMain.loadDocument();
 		});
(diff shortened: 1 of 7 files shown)


lib/DocumentService.php
```diff
@@ -142,17 +157,62 @@ public function getDocumentByUserId(string $userId, int $fileId, ?string $dir) :
 				return $this->reportError('Document for the fileId ' . $fileId . 'not found');
 			}
 
-			// Set basic parameters
+			/** @var \OCP\Files\Storage\IStorage $storage */
+			$storage = $document->getStorage();
+			$isSharedFile = $storage->instanceOfStorage('\OCA\Files_Sharing\SharedStorage');
+			$isFederatedShare = $storage->instanceOfStorage('\OCA\Files_Sharing\External\Storage');
```
jvillafanez, line 163: Any other alternative? Checking for specific classes doesn't look good

js/documents.js
```diff
@@ -487,7 +485,7 @@ var documentsMain = {
 			    "&lang=" + OC.getLocale().replace('_', '-') +
 			    "&closebutton=1" +
 			    "&revisionhistory=1";
-			if (!documentsMain.canEdit || action === "view") {
+			if (action === "view") {
```
mrow4a, line 488: this and other are really not needed nor used. actually "action" should determine how document should be opened. Then token is created with proper permissions, and if these are missing in the token worst case document instead of in edit opens in view mode.

lib/DocumentService.php
```diff
@@ -142,17 +157,62 @@ public function getDocumentByUserId(string $userId, int $fileId, ?string $dir) :
...
+					$ret['secureView'] = true;
+					$ret['secureViewId'] = $share->getId();
+				}
+			} elseif ($isFederatedShare) {
+				/** @var \OCA\Files_Sharing\External\Storage $storage */
+				/* @phan-suppress-next-line PhanUndeclaredMethod */
+				$ret['federatedServer'] = $storage->getRemote();
+				/* @phan-suppress-next-line PhanUndeclaredMethod */
+				$ret['federatedToken'] = $storage->getToken();
+			}
```
mrow4a, line 214: moved all code related to shares and permissions/attributes into DocumentService. Now DocumentController is not aware of this business logic related to UserFolder, external/share storages etc. It simplifies and isolates that complexity.

lib/Controller/DocumentController.php
```diff
@@ -174,58 +173,98 @@ public function index($fileId, $dir) {
...
+						$this->l10n->t('Collabora Online: Error encountered while opening the document.', []),
+						$this->l10n->t('Please contact the administrator.', [])
+					);
+				}
+
+				if (isset($docinfo['federatedServer'], $docinfo['federatedToken'])) {
+					return $this->responseError(
+						$this->l10n->t('Collabora Online: Federated shares not supported yet.', []),
+						$this->l10n->t('Please contact the administrator.', [])
+					);
```
mrow4a, line 194: this is the primary handling related to federates shares now.

mrow4a (commented):
@jvillafanez if you could have a look on this refactor.

lib/DocumentService.php
mrow4a: - it is public method of that class, so why not to use it?
- I had extensive look and this is the only way in OC currently, probably we would need to somehow handle this better in OC10 core?

jvillafanez: There could be other classes implementing a shared storage. I know there are other places where this kind of check is done, so maybe it's too late and switching implementations is no longer an option.

Anyway, if there is no better option, we'll have to go with this at least for now.

jvillafanez (approved):

Issue #381, opened 2021-02-16
Using Collabora Online via Federation


pmaier1:
**User story**
1. Use Collabora Online for viewing and editing documents collaboratively via federated sharing
2. Use Secure View sharing with Collabora via federation

**Further info**
- For 2.: Secure View via federation has to use the Collabora Server of the owner's instance (otherwise security is undermined!)

---

Needs technical assessment and concept. Please consider both separately. Both should work in the end but having only 1. first would be a good start.

@mrow4a

mrow4a:
@pmaier1 can we split it into two subproblems?
1. secure view with collabora via federation
- might be possibly due to read-only character of the feature
 
2. collabora online (edit etc) via federation 
- likely not possible, or if is, I sense problems

mrow4a:
@pmaier1 thanks for clarifying in the description, would consider and address challenges separately so we have understanding.

mrow4a:
# Technical Concept

## Use-case

- The ownCloud instances **sourcefedinstance** and **destfedinstance** are trusted entities with each other and thus federated shares can be created
- User **user1@sourcefedinstance** shares secure-view or document via federated sharing with **user1@destfedinstance**- As result of this action **fedsharetoken** gets created that permits **user1@destfedinstance** to access file of **user1@sourcefedinstance**, and **fedcollaboratoken** that permits access to **sourcefedinstance** collabora instance. 
- As **user1@destfedinstance** is not known to  **sourcefedinstance** as instance user, thus each access to the file is done via **user1@destfedinstance (on behalf of user1@sourcefedinstance)** (remote access).

## Option 1 - On open redirect to remote federated owncloud instance for access to source collabora
User opening document goes through federated collabora authentication flow (**fedsharetoken** and **collaboratoken**), and gets redirected directly to **sourcefedinstance** instance on specific path to access the collabora. 

- users accesses file from **sourcefedinstance** directly as result of redirect from **destfedinstance**
- as users gets fully redirected, user is under audit and governance of **sourcefedinstance**

### Technical flow:
-  **user1@destfedinstance** accessed federated share file by click "Open in Collabora"
- Request is sent to **destfedinstance** richdocuments app
- richdocuments app requests access token **fedcollaboratoken** from **sourcefedinstance** via **fedsharetoken** that gives access to file. File can only be accessed having valid pair of **fedcollaboratoken** and **fedsharetoken** .
[...]

PR #498, opened 2023-06-13
Federated shares support v1 

mrow4a:
implements https://github.com/owncloud/richdocuments/issues/381

- [x] on detected federated document redirect to remote server 
- [x] add basic "federated handshake" handling assigning "remote user" as editor uid
- [x] make sure to assign proper serverHost in the wopi token db entry on both servers (we need to keep track on both servers that the editing is happening and where the user got redirected)
- [x] make sure to assign proper access_token in the wopi token db entry on both servers (in case remotes need to exchange some further information between each other)
- [x] extend "federated handshake" to allow to retrieve editor uid and other wopi session details from remote
- [x] adjust tests
- [x] make sure to use new major version for the release @jnweiger

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -76,7 +76,8 @@ $.widget('oc.documentGrid', {
 	},
 
 	_load : function (fileId){
-		if (fileId){
+		var isRemoteFile = getURLParameter('shareToken');
+		if (fileId || isRemoteFile){
 			documentsMain.initSession();
 			return null;
 		}
--- a/js/documents.js
+++ b/js/documents.js
@@ -655,6 +655,9 @@ var documentsMain = {
 	onStartup: function() {
 		documentsMain.UI.init();
 
+		// TODO: needs invesitigation why comparison operator for null of return
+		// for getURLParamager need to be != 'null', ref. core/js/js.js 
+
 		// Does anything indicate that we need to autostart a session?
 		var fileId = getURLParameter('fileId');
 		if (fileId != 'null')
--- a/js/documents.js
+++ b/js/documents.js
@@ -76,8 +76,8 @@ $.widget('oc.documentGrid', {
 	},
 
 	_load : function (fileId){
-		var isRemoteFile = getURLParameter('shareToken');
-		if (fileId || isRemoteFile){
+		var isShare = getURLParameter('shareToken');
+		if (fileId || isShare) {
 			documentsMain.initSession();
 			return null;
 		}
--- a/js/documents.js
+++ b/js/documents.js
@@ -659,13 +659,22 @@ var documentsMain = {
 		var fileId = getURLParameter('fileId');
 		if (fileId != 'null')
 			documentsMain.fileId = fileId;
+
 		var dir = getURLParameter('dir');
 		if (dir != 'null')
 			documentsMain.returnToDir = dir;
-		var shareToken = getURLParameter('shareToken');
-		if (shareToken != 'null')
-			documentsMain.returnToShare = shareToken;
 
+		var shareToken = getURLParameter('shareToken');
+		if (shareToken != 'null') {
+			// check if local share or federated share
+			var server = getURLParameter('server');
+			if (server != 'null') {
+				documentsMain.returnToServer = server;
+			} else {
+				documentsMain.returnToShare = shareToken;
+			}
+		}
+		
 		// this will launch the document with given fileId
 		documentsMain.show(documentsMain.fileId);
 
@@ -828,6 +837,9 @@ var documentsMain = {
 		if (documentsMain.returnToDir) {
 			documentsMain.overlay.documentOverlay('show');
 			window.location = OC.generateUrl('apps/files?dir={dir}', {dir: documentsMain.returnToDir}, {escape: false});
+		} else if (documentsMain.returnToServer) {
+			documentsMain.overlay.documentOverlay('show');
+			window.location = documentsMain.returnToServer;
 		} else if (documentsMain.returnToShare) {
 			documentsMain.overlay.documentOverlay('show');
 			window.location = OC.generateUrl('s/{shareToken}', {shareToken: documentsMain.returnToShare}, {escape: false});
--- a/lib/Controller/DocumentController.php
+++ b/lib/Controller/DocumentController.php
@@ -26,6 +26,7 @@ use OCA\Richdocuments\Db;
 use OCA\Richdocuments\Db\Wopi;
 use OCA\Richdocuments\DiscoveryService;
 use OCA\Richdocuments\DocumentService;
+use OCA\Richdocuments\FederationService;
 use OCA\Richdocuments\Helper;
 use OCP\App\IAppManager;
 use OCP\AppFramework\Controller;
@@ -83,6 +84,11 @@ class DocumentController extends Controller {
 	private $discoveryService;
 
 	/**
+	 * @var FederationService The document service
+	 */
+	private $federationService;
+
+	/**
 	 * @var IGroupManager The group manager service
 	 */
 	private $groupManager;
@@ -120,7 +126,8 @@ class DocumentController extends Controller {
 		IGroupManager $groupManager,
 		IUserManager $userManager,
 		IPreview $previewManager,
-		INavigationManager $navigationManager
+		INavigationManager $navigationManager,
+		FederationService $federationService
 	) {
 		parent::__construct($appName, $request);
 		$this->l10n = $l10n;
@@ -134,6 +141,7 @@ class DocumentController extends Controller {
 		$this->userManager = $userManager;
 		$this->previewManager = $previewManager;
 		$this->navigationManager = $navigationManager;
+		$this->federationService = $federationService;
 	}
 
 	private function responseError($message, $hint = '') {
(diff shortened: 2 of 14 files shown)


mrow4a:
FYI @Ashod

lib/FederationService.php
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	/**
+	 * Get the wopiSrc Url from a remote server.
+	 *
+	 * @param string $server a remote
+	 * @return string with the wopi src Url
+	 */
+	public function getRemoteWopiSrc($server) {
+		if (\strpos($server, 'http://') === false && \strpos($server, 'https://') === false) {
```
jvillafanez, line 135: I'm not sure if it's possible, but if server is "abcd/efghttp://my.server.com" we won't be prefixing the protocol.
If we expect the server to have a specific format, for example, to be urlencoded, then it should be at least specified in the PHPDoc, otherwise I assume any string is valid.

lib/FederationService.php
```diff
@@ -0,0 +1,215 @@
...
+	 */
+	public function generateFederatedCloudID(string $userId) : string {
+		if (\strpos($userId, '@') === false) {
+			// generate federated cloud id
+			$user =  $userId;
+			$remote = \preg_replace('|^(.*?://)|', '', \rtrim($this->urlGenerator->getAbsoluteURL('/'), '/'));
+			return "{$user}@{$remote}";
+		}
+
+		// Find the first character that is not allowed in user names
```
jvillafanez, line 174: Too many little things here... maybe move this part to a specialized method so we can add unit tests?
In addition, smaller methods are easier to explain what they do.

lib/FileService.php
```diff
@@ -133,6 +133,55 @@ public function getFileHandle(int $fileId, string $ownerUID, string $userUID): ?
 		return null;
 	}
 
+	public function newFile(string $path, string $ownerUID, string $editorUID): ?File {
```
jvillafanez, line 136: If it's a public method, it should have a PHPDoc.

lib/FileService.php
```diff
@@ -133,6 +133,55 @@ public function getFileHandle(int $fileId, string $ownerUID, string $userUID): ?
...
+		// take 1st mount as here we access original version of the file from owner
+		$this->setupFS($ownerUID);
+
+		// create the folder first
+		if (!$this->rootFolder->nodeExists(\dirname($path))) {
+			$this->rootFolder->newFolder(\dirname($path));
+		}
+
+		// create a unique new file
+		$path = $this->rootFolder->getNonExistingName($path);
```
jvillafanez, line 179: I guess race conditions aren't worthy to check in this case. Any dangerous thing that could happen an case of a race condition?

lib/Controller/WopiController.php
```diff
@@ -122,28 +120,64 @@ public function __construct(
...
+		}
+
+		// get origin
+		$postMessageOrigin = $res['server_host'];
+
+		// get owner info
+		$ownerId = $res['owner'];
+
+		// get user info
+		if ($res['editor'] && $res['editor'] !== '' && !($res['attributes'] & WOPI::ATTR_FEDERATED)) {
```
jvillafanez, line 143: The `$res['editor'] && $res['editor'] !== ''` seems confusing. It seems `$res['editor']` is evaluated as a boolean even if it isn't one. This could backfire in special cases such as `$res['editor'] = '0'`

I also assume that the editor is mandatory, so in that case I don't know why we're checking `$res['editor']`

lib/Controller/WopiController.php
```diff
@@ -122,28 +120,64 @@ public function __construct(
...
+			// file editing as local logged in user
+			$editor = $this->userManager->get($res['editor']);
+
+			$userId = $editor->getUID();
+			$userFriendlyName = $editor->getDisplayName();
+			$userEmail = $editor->getEMailAddress();
```
jvillafanez, line 149: It might be useful to know that the `$userEmail` could be set to null if the user doesn't have an email set.

lib/Controller/WopiController.php
```diff
@@ -402,74 +424,51 @@ public function wopiPutFileRelative(string $documentId): JSONResponse {
...
 		if (!$file) {
-			$this->logger->warning('PutFileRelative: could not retrieve file', ['app' => $this->appName]);
+			$this->logger->warning('PutFileRelative: Could not retrieve file', ['app' => $this->appName]);
 			return new JSONResponse([], Http::STATUS_NOT_FOUND);
 		}
 
-		$path = '';
+		// Retrieve suggested target
+		$suggested = $this->request->getHeader('X-WOPI-SuggestedTarget');
+		$suggested = \iconv('utf-7', 'utf-8', $suggested);
```
jvillafanez, line 441: This might need a comment. It could be useful to know why we need to convert the charset because it isn't a common operation.

js/documents.js
```diff
@@ -658,13 +659,22 @@ var documentsMain = {
...
+
 		var dir = getURLParameter('dir');
 		if (dir != 'null')
 			documentsMain.returnToDir = dir;
-		var shareToken = getURLParameter('shareToken');
-		if (shareToken != 'null')
-			documentsMain.returnToShare = shareToken;
 
+		var shareToken = getURLParameter('shareToken');
+		if (shareToken != 'null') {
```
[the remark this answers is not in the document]
mrow4a: @jvillafanez I would prefer not to touch this, especially as you correctly noticed I am not sure about `getURLParameter`. Somehow this is used in here, and all returns from this check with this weird `!= 'null'` . I am a bit scared on touching it, but will include like TODO or FIXME.

lib/FederationService.php
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	/**
+	 * split user and remote from federated cloud id, null if not federated cloud id
+	 *
+	 * @param string $userId user id
+	 * @return string
+	 */
+	public function generateFederatedCloudID(string $userId) : string {
+		if (\strpos($userId, '@') === false) {
```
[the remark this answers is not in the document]
mrow4a: @jvillafanez this is copy of code from other federated codepaths. I would reference them here. These all, in case of changes, should be kept in sync. This logic is handled below more in detail.

lib/FederationService.php
mrow4a: it is already tested in different class/app. I could potentially used federation app class here with all these reflection stuff.. see above.

lib/FederationService.php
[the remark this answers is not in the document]
mrow4a: I can maybe see whether I can copy some tests..

js/documents.js
```diff
@@ -76,7 +76,8 @@ $.widget('oc.documentGrid', {
 	},
 
 	_load : function (fileId){
-		if (fileId){
+		var isShare = getURLParameter('shareToken');
```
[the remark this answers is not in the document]
mrow4a: ok, it is in `/var/www/owncloud/core/js/js.js`

lib/FileService.php
mrow4a: worked for many years, so hopefuly would work still . this is just move of a code to the dedicated class.

lib/Controller/WopiController.php
mrow4a: I would love to know too why this was done years back 😆

lib/Controller/WopiController.php
mrow4a: it is checked for null below.

lib/Controller/WopiController.php
mrow4a: I will add a comment

lib/Controller/WopiController.php
mrow4a: yes, you are right, editor is always at least empty string. I did not notice. thanks!

lib/FederationService.php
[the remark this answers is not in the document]
jvillafanez: If it's copied from another place maybe we can include a reference, specially if the checks should be in sync, otherwise I could touch this piece only here without knowing.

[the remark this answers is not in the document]
mrow4a: You were right here, because it contained logic that was in fact not really needed. when use clicks on federated share mount, owner is always federated user, and editor is always local user. Thus when generating token I just need to make editoruid being a federated user (as session is redirected to federated server)

lib/DocumentService.php
```diff
@@ -293,6 +304,66 @@ public function getDocumentByShareToken(string $token, ?int $fileId) : ?array {
...
+			$ret = [];
+			$ret['owner'] = $owner;
+			$ret['allowEdit'] = ($share->getPermissions() & Constants::PERMISSION_UPDATE) === Constants::PERMISSION_UPDATE;
+			$ret['allowExport'] = true;
+			$ret['allowPrint'] = true;
+			$ret['mimetype'] = $document->getMimeType();
+			$ret['path'] = $root->getRelativePath($document->getPath());
+			$ret['name'] = $document->getName();
+			$ret['fileid'] = $document->getId();
+			$ret['version'] = '0'; // latest
```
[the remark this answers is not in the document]
mrow4a: made it 0 (integer) now. Now it would be clear.

js/documents.js
```diff
@@ -658,13 +659,22 @@ var documentsMain = {
...
 			documentsMain.returnToDir = dir;
-		var shareToken = getURLParameter('shareToken');
-		if (shareToken != 'null')
-			documentsMain.returnToShare = shareToken;
 
+		var shareToken = getURLParameter('shareToken');
+		if (shareToken != 'null') {
+			// check if local share or federated share
+			var server = getURLParameter('server');
+			if (server != 'null') {
```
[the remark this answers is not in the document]
mrow4a: added comment

js/documents.js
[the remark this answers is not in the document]
mrow4a: added comment.

lib/Controller/DocumentController.php
```diff
@@ -304,24 +336,76 @@ public function public($shareToken, $fileId) {
...
-			$docinfo = $this->documentService->getDocumentByShareToken($shareToken, $fileId);
-			if (!$docinfo) {
-				$this->logger->warning("Cannot retrieve document from share {token} that has fileid {fileId}", ["token" => $shareToken, "fileId" => $fileId]);
-				return $this->responseError(
-					$this->l10n->t('Collabora Online: Error encountered while opening the document.', []),
-					$this->l10n->t('Please contact the administrator.', [])
-				);
-			}
+		// FIXME: In public links allow max 100MB
+		$maxUploadFilesize = 100000000;
```
jvillafanez, line 340: I'd prefer `100*1000*1000` since it's clearer. Too many zeros otherwise.

lib/Controller/DocumentController.php
```diff
@@ -332,7 +416,17 @@ public function public($shareToken, $fileId) {
 			);
 		}
 
-		// FIXME: In public links allow max 100MB
+		// Federated share is a user coming from remote instance so cannot show base template
+		$renderAs = 'empty';
```
jvillafanez, line 420: I'd move this var closer to where it's being used (line 454 I think).

lib/Controller/DocumentController.php
```diff
@@ -623,7 +721,7 @@ private function createWopiSessionForAuthUser(array $docInfo) : array {
 		$wopiSrc = $this->discoveryService->getWopiSrc($mimetype);
 		if (($allowEdit === true) && isset($wopiSrc['action'])
 				&& ($wopiSrc['action'] === 'edit' || $wopiSrc['action'] === 'view_comment')
-				&& ($this->isAllowedEditor($currentUser) === true) && ($version === '0')) {
+				&& ($this->isAllowedEditor($editorUid) === true) && ($version === 0)) {
```
jvillafanez, line 724: I guess `$version === 0` is intentional. Double-checking because now it's an integer, but it was a string before.

lib/Controller/WopiController.php
```diff
@@ -839,32 +846,41 @@ public function wopiUnlockAndRelock(string $documentId): JSONResponse {
...
+		$row = new Db\Wopi();
+		$res = $row->getWopiForToken($wopiToken);
+		if (!$res) {
+			$this->logger->debug('Cannot find token.', ['app' => $this->appName]);
+			return null;
 		}
-		return new JSONResponse([], Http::STATUS_OK);
+
+		// check if the token is for the given file
+		if ($res['fileid'] != $fileId) {
```
jvillafanez, line 879: Use `!==` or add a comment explaining why it's needed.

lib/DocumentService.php
```diff
@@ -293,6 +304,66 @@ public function getDocumentByShareToken(string $token, ?int $fileId) : ?array {
...
+	public function getDocumentByFederatedToken(string $token, ?string $path) : ?array {
+		try {
+			// Get share by token
+			$share = $this->shareManager->getShareByToken($token);
+			if (!$this->isShareAuthValid($share)) {
+				return null;
+			}
+
+			$node = $share->getNode();
+			if ($node->getType() == FileInfo::TYPE_FILE) {
```
jvillafanez, line 326: Use `===` or add a comment.

jvillafanez (approved):
Other than the `!==` and `===` which should be easy to fix, the rest is small stuff. I assume the logic works fins.

Issue #501, opened 2023-07-07
richdocuments 4.0.0 release


jnweiger:
This is the server app release template, please copy the part below in a new ticket.
For instructions around app releases, see https://confluence.owncloud.com/display/OG/Releasing+Apps

### Reason

- Release needed for
   - #498  
   - #497 
  - #496
  - #494 
  - #493 
  - #464 
  - [#493](https://github.com/owncloud/richdocuments/pull/493) - make sure to retrieve correct supershare based on current dir context
  - [#492](https://github.com/owncloud/richdocuments/pull/492) - Remove Symfony event dispatch from ignoreErrors
  - [#456](https://github.com/owncloud/richdocuments/pull/456) - web: add Drawing filetype and add it to + menu
  - [#467](https://github.com/owncloud/richdocuments/pull/467) - Set appropriate icon for web
  - [#462](https://github.com/owncloud/richdocuments/pull/462) - Adjust 'if' conditionals that were reported by phpstan
  - [#457](https://github.com/owncloud/richdocuments/pull/457) - typo fix (templates/documents.php)
  - [#455](https://github.com/owncloud/richdocuments/pull/455) - ensure ODG Drawing compatibility across integration
  - [#454](https://github.com/owncloud/richdocuments/pull/454) - Change Richdocuments app name to Collabora Online
  - [#451](https://github.com/owncloud/richdocuments/pull/451) - disable secure view js and settings when not available
  - [#468](https://github.com/owncloud/richdocuments/pull/468) - Upload button overlaps with a document icon in the second row
  - [#486](https://github.com/owncloud/richdocuments/pull/486) - E5515 feature/wopi locks

### Product approval 

- [ ] Verify transifex commits (enterprise should update weekly, others daily): If folder `l10n` exists, then `git log` should show '[tx] updated from transifex` ...
- [ ] Ping ProductManagement to check `info.xml` description and screenshot

### QA

- [x] Create **QA** Milestone and set the description to $version
- [ ] Move tickets from milestone `development` to `qa` (testplan and tickets mentioned in release ticket)
- [x] Create release branch with the format "release-$version"
- [x] Create "4.0.0 Testplan" --> #500  ([QA templates](https://github.com/owncloud/QA/tree/master/Server) or [Enterprise](https://github.com/owncloud/qa-enterprise/tree/master/Core))
- [ ] copy new/reopened issues from the previous release ticket as 'Issues to review' below.
[...]

PR #505, opened 2023-08-02
Zotero integration + refactor and bug fixes of admin/personal panel

mrow4a:
zotero required to add new personal panel to allow adding key. because of that I did not want to include more legacy logic, and I refactored the whole admin panel and fixed the bugs that were there. 

Screenshot 2023-08-02 at 08 17 09
Screenshot 2023-08-02 at 08 17 30

in case not enabled
Screenshot 2023-10-20 at 16 57 28


mrow4a:
@jvillafanez most of this logic is just migrated , with few fixes for some obvious issues like button no longer displaying the section or inconsistent logic between settings or actually wrong logic for saving user settings.

mrow4a:
@jnweiger would need translations in German (not sure how to organise this)

js/settings-personal.js
```diff
@@ -0,0 +1,77 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+(function ($, OC, OCA) {
+
+	OCA.RichdocumentsPersonalSettings = {
```
jvillafanez, line 22: It seems better to split the namespace into `OCA.Richdocuments.Settings.Personal`. At least I expect the name of the app to be after the `OCA` prefix. Whether the rest is `Settings.Personal`, `SettingsPersonal` or `PersonalSettings` is likely irrelevant and it might depend on future plans (I don't think we'll need to add too much js code into different namespaces.)

lib/Controller/SettingsController.php
```diff
@@ -46,47 +72,11 @@ public function list() {
...
-				'secure_view_has_watermark_default' => $this->appConfig->secureViewHasWatermarkDefaultEnabled() ? 'true' : 'false',
-				'secure_view_can_print_default' => $this->appConfig->secureViewCanPrintDefaultEnabled() ? 'true' : 'false',
-				'watermark_text' => $this->appConfig->getAppValue('watermark_text')
-			],
-			'blank'
-		);
-	}
-
-	public function update($wopi_url, $edit_groups, $doc_format, $test_wopi_url, $test_server_groups, $external_apps, $canonical_webroot, $menu_option, $secure_view_option, $secure_view_open_action_default, $secure_view_can_print_default, $secure_view_has_watermark_default, $watermark_text) {
+	public function setAdminSettings($wopi_url, $edit_groups, $doc_format, $test_wopi_url, $test_server_groups, $external_apps, $canonical_webroot, $menu_option, $secure_view_option, $secure_view_open_action_default, $secure_view_can_print_default, $secure_view_has_watermark_default, $watermark_text, $zotero) {
```
jvillafanez, line 79: definitely we need a different approach. Sending 14 different parameters seems terrible, and the next time we need an additional parameter I'm sure we'll send 15 (as you've done already)

In general, I'd prefer to use an array and document what keys are required and what are optional.

lib/Controller/SettingsController.php
```diff
@@ -142,14 +132,40 @@ public function update($wopi_url, $edit_groups, $doc_format, $test_wopi_url, $te
...
+		$uid = $this->userSession->getUser()->getUID();
+
+		if ($zoteroAPIPrivateKey !== null) {
+			$this->appConfig->setUserValue($uid, 'zoteroAPIPrivateKey', $zoteroAPIPrivateKey);
+		}
+
+		return new DataResponse(
+			[
+				'status' => 'success',
+				'data' => ['message' => (string) $message]
```
jvillafanez, line 167: Unless we expect specific information from the server, I think translations are better handled at the client's side. Just for a "saved" message, the client can check the status of the response and provide a translated message by itself.

templates/settings-personal.php
```diff
@@ -0,0 +1,36 @@
...
+	<p id="change_zotero_key_section-richdocuments" class="indent <?php if ($_['zotero'] !== 'true') {
+		p('hidden');
+	} ?>">
+		<br />
+		<label for="change_zotero_key-richdocuments"><?php p($l->t('Zotero Personal API Key:')) ?></label>
+		<input
+				type="text"
+				id="change_zotero_key-richdocuments" 
+				style="width: 350px; max-width: 100%"
+				value="<?php echo($_['zoteroAPIPrivateKey'] ? $_['zoteroAPIPrivateKey'] : ''); ?>"/>
```
jvillafanez, line 24: I'm not sure about security requirements on this one.

mrow4a:
@jvillafanez I think some docs will come from Collabora or ownCloud after we release it. Then we could update it here. 

but in general, admin allows users to provide keys and then Collabora would use them

jnweiger:
I'd target this for a 4.1.0 release. 4.0.0 is in RC-Phase and P2. I don't want to delay that any further.
I second the questions asked by @jvillafanez, 
In addition, I'd like to see instructions for a test setup. Server setup / keys ... demo data ...

Transifex requires that the code is in master. After that I can take care of translations, but that is not a priority currently.

mrow4a:
@jnweiger this is fine. But then I nedd to submit seperate patch for secure view. One feature (from settings) broke when I did refactor for federation. We better include that one in 4.0.0.

mrow4a:
lets finish and merge after release-4.0.0 is done in https://github.com/owncloud/richdocuments/issues/501

js/settings-admin.js
```diff
@@ -0,0 +1,333 @@
...
+				);
+			});
+		},
+
+		setAdminSettings: function(data, doneCallback, failCallback) {
+			$.post(
+				OC.generateUrl("apps/richdocuments/ajax/settings/setAdminSettings"),
+				data
+			).done(function (response) {
+				doneCallback(response);
```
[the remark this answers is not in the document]
mrow4a: done, makes sense

lib/Controller/SettingsController.php
mrow4a: done, makes sense

js/settings-personal.js
mrow4a: done, makes sense

templates/settings-personal.php
mrow4a: addressed it with the way we currently handle it in OC

lib/Controller/SettingsController.php
mrow4a: this is something I would prefer not to touch.

lib/Panels/Personal.php
```diff
@@ -0,0 +1,73 @@
...
+		$uid = $this->userSession->getUser()->getUID();
+
+		$zoteroEnabled = $this->appConfig->getAppValue('zotero');
+		$zoteroAPIPrivateKey = $this->appConfig->getUserValue($uid, 'zoteroAPIPrivateKey');
+
+		$template = new Template('richdocuments', 'settings-personal');
+
+		$template->assign('zotero', $zoteroEnabled);
+		if ($zoteroAPIPrivateKey !== null) {
+			$template->assign('zoteroAPIPrivateKey', $zoteroAPIPrivateKey);
```
jvillafanez, line 69: From a security point of view, is it secure to send the private API key? We could just send a "private key is set" message instead. If needed, we could include a delete button to remove the API key.

lib/Controller/SettingsController.php
jvillafanez: Maybe include a TODO comment?

mrow4a: good point, done

jvillafanez (approved):

jnweiger:
German translations are at 100% currently, including zotero related strings.
Praise to @EParzefall and Community translators.

Issue #514, opened 2023-09-16
'Office' view does not work any more with 4.0.0


timar:
I installed the pre-release 4.0.0 to two demo sites. The 'Office' view worked on neither of them. ownCloud version	10.12.2.1 and 10.13.1 (production)

Click on Office. 
![image](https://github.com/owncloud/richdocuments/assets/290614/33f99755-6d94-4d2f-9e27-32e37cbb2737)

Expected: something like this (screenshot from 3.0.0):
![image](https://github.com/owncloud/richdocuments/assets/290614/e1070f58-3629-4ff7-994d-16e83f83e1ec)

Actual: ownCloud displayed 'Files' view.

jnweiger:
Yes, same issue here. A blank page appears. And the Javascript console has an error:

![grafik](https://github.com/owncloud/richdocuments/assets/1108546/4440f132-46f7-44b6-a604-9eb230d34e69)

jnweiger:
https://github.com/owncloud/richdocuments/blob/master/js/documents.js#L732-L741
```
var path = $('li[data-fileid='+ id +']>a').attr('href');
```
I don't understand much of that javascript. I that some dom-tree expression syntax? It seems that `id` is empty. 
@mrow4a is that an issue?

mrow4a:
I do not want to be too pesimistic but can be owncloud core broke this.

jnweiger:
Not reproducable with richdocuments-3.0.0 on the latest greatest 10.13.1
Reproduced with richdocuments-4.0.0 on very old 10.11.0

mrow4a:
hmm ok, I think I know what is the issue.. with federated shares probably did not notice it breaks other pannel.

mrow4a:
fixed in https://github.com/owncloud/richdocuments/pull/517

jnweiger:
Confirmed, #517 fixes the issue! Thanks Piotr!

jnweiger:
Fixed in https://github.com/owncloud/richdocuments/releases/download/v4.0.0-rc.5/richdocuments-4.0.0-rc.5.tar.gz

PR #517, opened 2023-09-22
fix issue with null return

mrow4a:
@jvillafanez at least now we know why the comparison operator was needed..... fixes https://github.com/owncloud/richdocuments/issues/514


jnweiger (approved):
Yes, that fixes the issue perfectly fine!
(My own research went in exactly the same direction)

Issue #415, opened 2021-07-15
Revision History broken


theheyon:
Hey on there!

Revision History seems to be broken ((

![image](https://user-images.githubusercontent.com/79450887/125823094-bd4c6845-ddc9-48d9-99fd-fa41ce76895e.png)

System:
CODE running on Ubuntu 18.04.4 LTS
LOOLWSD 6.4.9
LOKit Collabora Office 6.4-39
Collabora Online app 2.5.0
ownCloud 10.7.0

Same issues here:
https://github.com/owncloud/richdocuments/issues/275
https://github.com/owncloud/richdocuments/issues/222

Maybe, a way to fix from NC:
https://github.com/nextcloud/richdocuments/issues/9
https://github.com/nextcloud/richdocuments/pull/27

theheyon:
/js/documents.js:

![image](https://user-images.githubusercontent.com/79450887/126013590-be9a689e-d336-4639-b566-ce5d25cdf6a4.png)

maybe, someone forgot about this? )
or OC10x still not supports a feature?

theheyon:
maybe, versioning API could be used instead files_versions app itself?

https://doc.owncloud.com/server/developer_manual/webdav_api/files_versions.html

mrow4a:
This is not an easy fix @hodyroff . The issue here is that we switched for version to META endpoint and DAV (Propfind/Copy/etc). This means, it would be impossible to access it that way from frontend. 

However, what can be done is what onlyoffice did. Backend integration. We just put a wrapper on top of oc10 versions, and access them directly. Thus implement our own controller here. The limitation would be that we probably could be only seeing contents of versions, but we cannot manage versions (restore/publish/dowload/etc), this would need to be done from versions tab in OC UI. Can probably be done via some button to redirect there 🤷🏻‍♂️ 

I think this is good solution as streamlines with OnlyOffice, and I would progress with it.

Issue #222, opened 2018-10-02
Revision history and versioning don't work


HeberLoran:
I just made the installation of the plugin (2.0.8) in version 10.0.14 of ownclud. I can edit and save the changes of the documents in collabora without any problem, but in owncloud it is not saved who modified the document, it is only modified by "remote user", in ouwncloud if you see the version history, but when opening the file in collabora no.
Does anyone know how I can correct this detail?

![version_collabora](https://user-images.githubusercontent.com/43792891/46369289-2249b180-c648-11e8-80a2-3b1c129b858d.JPG)
![activity_collabora](https://user-images.githubusercontent.com/43792891/46369290-2249b180-c648-11e8-90f2-7a496a03e36f.JPG)

fgarpe:
I have the same issue, this problem breaks the traceability in document management.

LukaszSellmann:
I have this problem too . It's still not resolved in version 2.3.0 of richdocuments app.

PR #522, opened 2023-09-27
Bugfix: broken version revision

mrow4a:
fixes: https://github.com/owncloud/richdocuments/issues/415 , https://github.com/owncloud/richdocuments/issues/222

I attempted to do it in such a way that it does not affect code complexity and performance of the "non-versioned" code paths. Thus performance is affected only when versions are listed or accessed

Screenshot 2023-09-27 at 15 58 30

Screenshot 2023-09-27 at 15 58 42

### DIFF
--- a/js/documents.js
+++ b/js/documents.js
@@ -330,11 +330,8 @@ var documentsMain = {
 				downloadUrl = OC.generateUrl('apps/files/download'+ documentPath);
 			} else {
 				// FIXME: this is no longer supported in OC10
-				downloadUrl = OC.generateUrl('apps/files_versions/download.php?file={file}&revision={revision}',
-				                             {file: documentPath, revision: version});
-				fileId = fileId + '_' + version;
-				restoreUrl = OC.generateUrl('apps/files_versions/ajax/rollbackVersion.php?file={file}&revision={revision}',
-				                            {file: documentPath, revision: version});
+				downloadUrl = '';
+				restoreUrl = '';
 			}
 
 			var revHistoryItemTemplate = Handlebars.compile(documentsMain.UI.revHistoryItemTemplate);
--- a/js/documents.js
+++ b/js/documents.js
@@ -245,20 +245,14 @@ var documentsMain = {
 			'<div id="revisionsContainer" class="loleaflet-font">' +
 			'<ul></ul>' +
 			'</div>' +
-			'<input type="button" id="show-more-versions" class="loleaflet-font" value="{{moreVersionsLabel}}" />' +
 			'</div>',
 
 		revHistoryItemTemplate: '<li>' +
-			'<a href="{{downloadUrl}}" class="downloadVersion has-tooltip" title="' + t('richdocuments', 'Download this revision') + '"><img src="{{downloadIconUrl}}" />' +
 			'<a class="versionPreview"><span class="versiondate has-tooltip" title="{{formattedTimestamp}}">{{relativeTimestamp}}</span></a>' +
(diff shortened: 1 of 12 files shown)


mrow4a:
fyi @hodyroff @pako81 that the issue is addressed here. not sure which enterprise issue this is

js/documents.js
```diff
@@ -353,54 +336,49 @@ var documentsMain = {
...
+			}).always(function(){
+				// show revisions container after loading
+				documentsMain.overlay.documentOverlay('hide');
+				$('#revPanelContainer').show();
 
-			// add "Show more versions" button
-			$('#show-more-versions').click(function(e) {
-				e.preventDefault();
-				documentsMain.UI.fetchAndFillRevisions(documentPath);
+				// fake click on first revision (i.e current revision)
```
jvillafanez, line 378: what if you have multiple revisions and you're on the third one? Not sure when the `showRevHistory` is called, but going back to the first revision if you just want to show the revisions doesn't look a good behavior.

lib/Controller/DocumentRevisionController.php
```diff
@@ -0,0 +1,167 @@
...
+	 * Get collabora document revisions for:
+	 * - the base template if fileId is null
+	 * - file in user folder (also shared by user/group) if fileId not null
+	 *
+	 * @NoAdminRequired
+	 */
+	public function list($fileId) {
+		if (\is_numeric($fileId)) {
+			// parse fileId pointing to file
+			$fileId = (int) $fileId;
```
jvillafanez, line 64: we might need to check the behavior with unreasonable big numbers. It might be ok if we document that numbers bigger than 2⁶⁴ will be considered as the maximum integer value (if that's the behavior)
As long as the behavior is ok, the documentation should be enough.

lib/Controller/DocumentRevisionController.php
```diff
@@ -0,0 +1,167 @@
...
+		IUserSession $userSession,
+		IRootFolder $rootFolder
+	) {
+		parent::__construct($appName, $request);
+		$this->userSession = $userSession;
+		$this->rootFolder = $rootFolder;
+	}
+
+	/**
+	 * Get collabora document revisions for:
```
jvillafanez, line 55: Not sure if it's worthy to say that the current (latest) version of the file will be excluded (or so it seems).

jvillafanez:
We might need to add some comment somewhere about the version "0". It seems counter-intuitive to me, because if the version 0 is the latest, then version 1 should be the second latest and version X the first version.
If the version 0 is special to mark the current file, but the rest of versions follows the expected order (1 for the first, 2 for the second, and so on), we should document it somewhere to avoid confusion.

lib/Controller/DocumentRevisionController.php
mrow4a: we have this handling everywhere in WOPI. lets keep it like is, I know suboptimal

js/documents.js
mrow4a: This is how collabora works, a bit weird. if this fake click is not done, you would have blank screen and then would need to select version.

lib/Controller/DocumentRevisionController.php
mrow4a: done

mrow4a:
@jvillafanez as requested, added a bit of clarification on why 0 is used. it is legacy of course. Because somehow in the past it was decided this column is not-null 😮‍💨

jvillafanez (approved):

PR #531, opened 2023-11-16
fix: drop JavaScript function `tipsy()` in favor of bootstrap tooltip

DeepDiver1975:


phil-davis:
Are we merging this and making patch release?

pako81:
> Are we merging this and making patch release?

Yes, it needs to go out before 10.14.0 as richdocuments app is not bundled.

jnweiger:
Yes. please approve this, I'll build richdocuments 4.3.0 with this.
Timing is relaxed now:  tipsy() support is back. -- release is before 10.15.0

PR #623, opened 2026-09-10
test: add a javascript unit test harness for the classic frontend

oc-tmueller:
## Summary

The scripts in `js/` had no automated coverage and CI had no JavaScript job at
all, so changes to the classic frontend could only be verified by hand. This
adds the setup the other ownCloud apps use - karma with jasmine, specs in
`tests/js/`, a `make test-js` target, and the shared `js-unit` reusable
workflow.

- `tests/js/karma.config.cjs` loads jQuery, jQuery UI and `OC` from the
  surrounding core checkout the same way the server loads them, so the scripts
  under test run unmodified. Only `js/documents.js` is loaded: the other scripts
  in `js/` are never on the same page and share one global namespace, which is
  also why core's own karma config lists files explicitly instead of globbing.
  The file is `.cjs` because `package.json` declares `"type": "module"`.
- `karma-jasmine-sinon` is needed because core's `specHelper.js` installs a
  global `beforeEach` that uses `sinon.fakeServer`. That also means no spec can
  reach the network, and each spec gets a `#testArea` div for fixtures.
- The `plugins` are listed explicitly. Karma's default `karma-*` discovery only
  scans the directory karma itself sits in, which finds nothing in a pnpm
  layout.
- The `test-js` target replaces the previous npm and bower rules in the
  `Makefile`. Those were dead: none of `nodejs_deps`, `bower_deps`, `BOWER` or
  `JSDOC` was ever defined, so the rule applied to an empty target. `make dist`
  output is unchanged.
- Node dependencies install with pnpm 10 through `npx`, since CI has node but
  not necessarily pnpm. Override with `PNPM=<command>` to use a local one. The
  new devDependencies were resolved into the lockfile #624 regenerated, so the
  only lockfile changes here are the karma, jasmine and sinon additions.

## Testing

37 specs in `tests/js/documentsSpec.js`, covering the frontend half of #620 -
which had to be verified manually at the time, for lack of this harness:

- `_absoluteHttpUrl()` refuses everything that is not an absolute http(s) URL,
  and accepts a subdirectory install and a non-default port.
- `_wopiOrigin()` derives the origin from an absolute and from a
  server-relative discovery `urlsrc`, and has no origin at all for an empty or
  non-http(s) one.
- `WOPIPostMessage()` posts to that origin rather than to any origin, and posts
  nothing when the origin is unknown.
- `onStartup()` takes the return URL from the hidden input the server renders
  and ignores a `server` URL parameter.
[...]

PR #634, opened 2026-09-16
build: drop the appstore target in favour of dist

### LINKED ISSUE #622: build: the appstore make target duplicates dist and its exclude list has drifted
The `Makefile` has two packaging targets that have diverged. `dist` uses an allowlist, `appstore` uses a denylist, and only the denylist has rotted.

## What is actually released

Releases are unaffected. `.github/workflows/release.yml` sets

```yaml
artifact-glob: build/dist/richdocuments.tar.gz
```

and `owncloud/reusable-workflows/.github/workflows/release.yml` runs `make dist`. `dist` copies only

```make
doc_files=README.md
src_dirs=appinfo assets css img js l10n lib templates
src_files=admin.php settings.php
all_src=$(src_dirs) $(src_files) $(doc_files)
```

so published tarballs contain exactly the intended files. `appstore` writes elsewhere anyway (`build/richdocuments.tar.gz`, not `build/dist/`), so nothing in CI consumes it.

## The problem

`appstore` rsyncs the whole project tree and subtracts a hand-maintained exclude list. That list has not kept up with the repository. Everything below is currently packaged by `make appstore` and is not part of `all_src`:

```
.eslintrc.json          .github                 .php-cs-fixer.dist.php
.prettierrc.json        .tsconfig.json          .well-known
AGENTS.md               CHANGELOG.md            CLAUDE.md
CODE_OF_CONDUCT.md      SECURITY.md             SUPPORT.md
package.json            pnpm-lock.yaml          sonar-project.properties
src                     tsconfig.json           vite.config.ts
```

That includes frontend sources (`src/`), the whole `.github/` directory, and the community health files added in #601.

Two individual signs of drift worth calling out:

- the list excludes `.php_cs.dist`, a file that no longer exists; the current config is `.php-cs-fixer.dist.php`, which is therefore packaged
- it excludes `CONTRIBUTING.md` but none of the sibling health files, so the intent was clearly to keep root documentation out

`appstore` also calls `occ integrity:sign-app` unconditionally against `$(cert_dir)`, so it fails outright without release certificates present locally.

## Suggested fix

Rather than adding another dozen `--exclude` lines, drop the denylist. Either delete `appstore` outright, since nothing uses it and `dist` supersedes it, or reduce it to a thin alias for `dist`. An allowlist cannot silently pick up new root files the way this one just did.

Low priority: no released artefact is wrong today. Filing it so the duplicate target does not quietly become correct-looking again.

oc-tmueller:
Closes #622.

The `Makefile` had two packaging targets. `dist` copies an explicit allowlist
(`all_src`) and is what ships — `release.yml` points `artifact-glob` at
`build/dist/richdocuments.tar.gz` and the reusable release workflow runs
`make dist`. `appstore` rsynced the whole working tree and subtracted a
hand-maintained exclude list, writing to `build/richdocuments.tar.gz`, which
nothing reads.

Only the denylist rotted. It now lets through **19 tracked root entries** that are
not part of `all_src`:

```
.eslintrc.json  .github         .php-cs-fixer.dist.php  .prettierrc.json
.tsconfig.json  .well-known     AGENTS.md               CHANGELOG.md
CLAUDE.md       CODE_OF_CONDUCT.md  SECURITY.md         SUPPORT.md
package.json    pnpm-lock.yaml  pnpm-workspace.yaml     sonar-project.properties
src             tsconfig.json   vite.config.ts
```

Two things the issue did not capture:

- it excluded `vendor/bin` and `vendor-bin` but **never `vendor` or `node_modules`**, so after
  any build it packaged those too
- `occ=` was assigned **twice**; the later `occ=$(CURDIR)/../../occ` already overrode the
  earlier `../core/occ`, which for an app at `apps/<app>/` resolves to `apps/core/occ` and
  never existed

Five of its excludes also name files that no longer exist (`.drone.star`, `.php_cs.dist`,
`.scrutinizer.yml`, `nbproject`, `screenshots`), and it called `occ integrity:sign-app`
unconditionally, so it failed outright without release certificates present.

### Why delete rather than extend the exclude list

An allowlist cannot silently pick up new root files the way this one just did —
`pnpm-workspace.yaml` joined the list days after the issue was filed. `activity`, `oauth2`
and `customgroups` already have no such target, and the reusable release workflow now asserts
artifacts are free of development files, so an appstore-style tarball would fail that gate
anyway.

Removed with it, all dead once the target is gone: `project_dir`, `sign_dir`, `appstore_dir`,
`source_dir`, `package_name`, the duplicate `occ=`, and the two references to `$(bower_deps)`
— a variable undefined anywhere since the bower rules went in #623.

### Verification

The released artifact is unchanged, measured rather than argued. In `owncloudci/php:8.3`
(GNU Make 4.3), building both this commit and its parent:

| check | result |
|---|---|
| `make dist` tarball | **171 entries, listings byte-identical** |
| `make help` | identical |
| `make appstore` before | failed `Makefile:34 Error 127` — the unconditional signing step |
| `make appstore` after | `No rule to make target 'appstore'` |

Staging `appstore` on the parent commit confirmed the drift concretely: all 19 entries above
were present in `build/sign/richdocuments/`, plus `node_modules/` and `vendor/`.

Backport to `4.2` follows separately — the `Makefile` is byte-identical between the branches.


phil-davis (approved):
Good, get rid of crud and just have `make dist`  for building a release.
Stuff like this hangs around for years, probably lots of core apps need their Make files cleaned up.

